PR #1, opened 2017-07-16
Fix Transfer Command

UnseenFaith:
This changes transfer so that it passes an array to ...join, instead of the String being concat into
` S\y\s\t\e\m\c\o\m\m\a\n\d\j\s`

Also fixes 'Reloaded undefined' `piece.name -> piece.help.name`

Removed the unnecessary .catch() because .load() is not a promise anymore.

You should consider doing `require("klasa");` instead of `require('index');` because when users transfer they will get errors when trying to load the new command because index is not defined in the local directory, thus making it seem like the transferred failed even if it didn't. Requiring klasa inside commands will fix this.

### DIFF
--- a/src/commands/System/transfer.js
+++ b/src/commands/System/transfer.js
@@ -23,7 +23,7 @@ module.exports = class extends Command {
 			})
 			.catch((err) => {
 				this.client.emit('error', err.stack);
-				return msg.sendMessage(`Transfer of ${piece.type}: ${piece.name} to Client has failed. Please check your Console.`);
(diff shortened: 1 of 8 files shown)


UnseenFaith:
WIP, realized the rest of theme are also broken/not working.

UnseenFaith:
Added Delete methods to all stores except Extendables
Fixed CommandLogging name
Made the stores the write class names to not confuse anyone
Made all load functions return a file array so they are all the same, which will allow it to work with transfer

PR #8, opened 2017-08-04
Simplify permission levels/structure

dada1134:
Changes:

- `KlasaClient.permStructure` gets replaced with `KlasaClient.permLevels`.
- The `PermissionLevels` class has been simplified. It is now a `Map` itself.
- The amount of `permLevel`s can now be configured when initializing the `PermissionLevels` class, but if not specified the default is still 11: `new PermissionLevels(15)`.

New way to get a `permLevel`: `client.permLevels.get(level)`.
It's also possible to overwrite a level now. If that's bad tell me.

### DIFF
--- a/src/lib/client.js
+++ b/src/lib/client.js
@@ -131,9 +131,9 @@ class KlasaClient extends Discord.Client {
 
 		/**
 		 * The permissions structure for this bot
-		 * @type {validPermStructure}
+		 * @type {PermissionLevels}
 		 */
-		this.permStructure = this.validatePermStructure();
+		this.permLevels = this.validatePermLevels();
 
 		/**
 		 * The threshold for how old command messages can be before sweeping since the last edit in seconds
@@ -205,17 +205,13 @@ class KlasaClient extends Discord.Client {
 	/**
 	 * Validates the permission structure passed to the client
 	 * @private
-	 * @returns {validPermStructure}
+	 * @returns {PermissionLevels}
 	 */
-	validatePermStructure() {
-		const structure = this.config.permStructure instanceof PermLevels ? this.config.permStructure.structure : null;
-		const permStructure = structure || this.config.permStructure || KlasaClient.defaultPermStructure.structure;
-		if (!Array.isArray(permStructure)) throw 'PermStructure must be an array.';
-		if (permStructure.some(perm => typeof perm !== 'object' || typeof perm.check !== 'function' || typeof perm.break !== 'boolean')) {
-			throw 'Perms must be an object with a check function and a break boolean.';
-		}
-		if (permStructure.length !== 11) throw 'Permissions 0-10 must all be defined.';
-		return permStructure;
+	validatePermLevels() {
+		const permLevels = this.config.permsLevels || KlasaClient.defaultPermLevels;
+		if (!(permLevels instanceof PermLevels)) throw new Error('permLevels must be an instance of the PermissionLevels class');
+		if (permLevels.isValid()) return permLevels;
+		throw new Error('Unexpected catastrophic failure at permission level validation!');
 	}
 
 	/**
@@ -321,7 +317,7 @@ class KlasaClient extends Discord.Client {
  * The default PermLevels
  * @type {PermissionLevels}
  */
-KlasaClient.defaultPermStructure = new PermLevels()
+KlasaClient.defaultPermLevels = new PermLevels()
 	.addLevel(0, false, () => true)
 	.addLevel(2, false, (client, msg) => {
 		if (!msg.guild || !msg.guild.settings.modRole) return false;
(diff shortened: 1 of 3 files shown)


src/lib/util/permLevels.js
```diff
@@ -33,26 +43,28 @@ class PermissionLevels {
...
+				typeof level.check !== 'function' ? throwE(`"check" in permission level ${index} must be a function`)
+				                                  : true;
+		});
+		/* eslint-enable no-multi-spaces, arrow-body-style, operator-linebreak, indent, no-mixed-spaces-and-tabs */
 	}
 
 }
 
+const throwE = msg => { throw new Error(msg); };
+const every = iterable => func => Array.prototype.every.call(iterable, func);
```
bdistin: Use this instead: https://github.com/hydrabolt/discord.js/blob/stable/src/util/Collection.js#L279-L292

src/lib/util/permLevels.js
```diff
@@ -33,26 +43,28 @@ class PermissionLevels {
...
-		for (let i = 0; i < 11; i++) {
-			const myLevel = this.levels.get(i);
-			if (myLevel) structure.push(myLevel);
-			else structure.push({ break: false, check: () => false });
-		}
-		return structure;
+	isValid() {
+		/* eslint-disable no-multi-spaces, arrow-body-style, operator-linebreak, indent, no-mixed-spaces-and-tabs */
+		return every(this)((level, index) => {
+			// Trick to replace a switch case or if else. Condition on the left, what's executed on the right. Last one is default.
```
bdistin: isValid should always return a boolean: Please use standard ```javascript
if (condition) return false;
if (condition) return false;
return true;
```
additionally you could add a debug method, to return a string telling you what entry isn't valid.

File: src/lib/util/permLevels.js [hunk 3]
```diff
@@ -15,14 +11,28 @@ class PermissionLevels {
...
+	static get defaultRequiredLevels() {
+		return 11;
+	}
+
+	/**
 	 * Creates a new PermissionLevels
+	 * @param {number} levels How many permission levels there should be
 	 */
-	constructor() {
+	constructor(levels = PermissionLevels.defaultRequiredLevels) {
```
bdistin: just do `= 11` remove the static getter

src/lib/client.js
```diff
@@ -205,17 +205,13 @@ class KlasaClient extends Discord.Client {
...
-		if (permStructure.some(perm => typeof perm !== 'object' || typeof perm.check !== 'function' || typeof perm.break !== 'boolean')) {
-			throw 'Perms must be an object with a check function and a break boolean.';
-		}
-		if (permStructure.length !== 11) throw 'Permissions 0-10 must all be defined.';
-		return permStructure;
+	validatePermLevels() {
+		const permLevels = this.config.permsLevels || KlasaClient.defaultPermLevels;
+		if (!(permLevels instanceof PermLevels)) throw new Error('permLevels must be an instance of the PermissionLevels class');
+		if (permLevels.isValid()) return permLevels;
+		throw new Error('Unexpected catastrophic failure at permission level validation!');
```
bdistin:
```javascript
throw new Error(permLevels.debug());
```

src/lib/util/permLevels.js
```diff
@@ -51,20 +44,49 @@ class PermissionLevels extends Map {
...
 
-}
+	/**
+	 * Checks if all permission levels are valid
+	 * @return {string} Error message(s)
+	 */
+	debug() {
+		const errors = [];
+		this.forEach((level, index) => {
+			if (typeof level !== 'object') {
```
bdistin: single line if possible (I should look for an eslint rule for this...)

src/lib/util/permLevels.js
```diff
@@ -51,20 +44,49 @@ class PermissionLevels extends Map {
...
 	}
 
-}
+	/**
+	 * Checks if all permission levels are valid
+	 * @return {string} Error message(s)
+	 */
+	debug() {
+		const errors = [];
+		this.forEach((level, index) => {
```
bdistin: use for of loop here

src/lib/util/permLevels.js
```diff
@@ -51,20 +44,49 @@ class PermissionLevels extends Map {
...
 	isValid() {
-		/* eslint-disable no-multi-spaces, arrow-body-style, operator-linebreak, indent, no-mixed-spaces-and-tabs */
-		return every(this)((level, index) => {
-			// Trick to replace a switch case or if else. Condition on the left, what's executed on the right. Last one is default.
-			return   typeof level !== 'object'   ? throwE(`Permission level ${index} must be an object`) :
-				typeof level.break !== 'boolean'  ? throwE(`"break" in permission level ${index} must be a boolean`) :
-				typeof level.check !== 'function' ? throwE(`"check" in permission level ${index} must be a function`)
-				                                  : true;
+		return this.every(level => {
+			if (typeof level !== 'object') return false;
```
bdistin: I edited my original comment... should be:
```javascript
return this.every(level => typeof level === 'object' && typeof level.break === 'boolean' && typeof level.check === 'function');
```

PR #18, opened 2017-09-05
Added readyMessage as a KlasaClient config option

MrJacz:
From the help from HellPie and AoDude, I here now pronounce you can change Klasa's ready message 💯


MrJacz:
False: Errors(cant split a boolean)
null: returns no ready msg
function: with client passed works great
string: works
undefined: passes the default message (function)

MrJacz:
^ With Ao's code cause, i didn't want to do if/else if(cause I didn't think it was needed smh)

MrJacz:
and tutorial is done

PR #19, opened 2017-09-06
Redesigned log output

HellPie:
- Added `wtf` (What a Terrible Failure, according to Android's Log class) and `verbose` log levels
- Added event shortcuts for the new levels (`wtf` is disabled by default to avoid usage due to its nature)

### Refactoring
- Log formatting is now managed through `this.format` in [log.js](https://github.com/HellPie/klasa/blob/log-update/src/events/log.js#L13-L20).
- Console method to use is also managed there.

### Restyling
- Every log level except for `error`, `warn` and `wtf` only colors the text, not the background, to better show warns and errors.
- Level `verbose` is colored grey (chalk `dim` does not work in most Windows environments) to not spam too many colors for those who will need to read the logs
- Level `wtf` is marked as underlined and bold, and highlights the entire message to be instantly recognizable due to the fact it should only be used for extreme case scenarios and "Terrible Failures".

### Screenshot testing how this pr will change logs:
image

### DIFF
--- a/src/events/log.js
+++ b/src/events/log.js
@@ -29,7 +29,7 @@ module.exports = class extends Event {
 
 		const format = this.formats[type || 'log'];
 		let timestamp = this.client.config.disableLogTimestamps ? '' : `[${moment().format('YYYY-MM-DD HH:mm:ss')}]`;
-		if ('time' in format) timestamp = format.type(timestamp);
+		if ('time' in format) timestamp = format.time(timestamp);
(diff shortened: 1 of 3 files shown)


MrJacz:
👍

CyberiumShadow:
Feel free to port this to Komada if you feel like it

HellPie:
ported quickly to komada, should be the same as here, mostly changed the code style

PR #20, opened 2017-09-07
Added botPerms to every system command and updated Events tutorial.

MrJacz:
As seen is this pr, every system cmd built into Klasa has been updated with botPerms and have updated the events tutorial please feel free to ask me any questions.


bdistin:
smh, the command handler itself will not process very far if the bot cannot send messages in the context it is in... send messages as a botPerm doesn't need to be set ever...

PR #22, opened 2017-09-11
for unhandledRejection added this.emit()

MrJacz:


kyranet:
Now, I'll give you something to think of. What's `this` in that function?

MrJacz:
client

UnseenFaith:
What the's trying to say is that this would not represent client in that function. It would most likely represent process or the listener for that event

MrJacz:
I know i didnt look 😃

bdistin:
Actually, it's an arrow function, so `this` should be KlasaClient the class, not the instance... Unfortunately that doesn't change that it still wouldn't work. This being in klasa is actually bad practice, but it's there to be a crutch for d.js-dev and klasa-dev until they are both stable enough the crutch can be removed safely.

PR #26, opened 2017-09-19
Typings (Second attempt, damn you CRLF)

kyranet:


src/lib/structures/Finalizer.js
```diff
@@ -18,7 +18,7 @@ class Finalizer {
 	/**
 	 * @param {KlasaClient} client The Klasa Client
 	 * @param {string} dir The path to the core or user finalizer pieces folder
-	 * @param {Array} file The path from the pieces folder to the finalizer file
+	 * @param {string[]} file The path from the pieces folder to the finalizer file
```
bdistin: should just be a string here.

kyranet: You're right, in the typings, I wrote

```typescript
public file: string;
```

But wrote the wrong type in the JSDocs (got confused with Command). Fixed 👍

PR #29, opened 2017-09-20
Added some exports, little cleanup

kyranet:
Fixes... real fixes. Lemme test this before, por favor.


kyranet:
overloads* Should be fine to merge now.

PR #40, opened 2017-10-02
reaction options enhancement

pedall:
- Replace default emojis instead of `Object.assign` in `setEmojis()` method (shouldn't this method be static)
- add possibility to remove reaction options from default by adapting `_determineEmojis` to only push existing emojis from this.emojis to the array

### DIFF
--- a/src/lib/util/RichDisplay.js
+++ b/src/lib/util/RichDisplay.js
@@ -128,7 +128,12 @@ class RichDisplay {
 	async run(msg, options = {}) {
 		if (!this.footered) this._footer();
 		if (!options.filter) options.filter = () => true;
-		const emojis = this._determineEmojis([], !('stop' in options) || ('stop' in options && options.stop), !('jump' in options) || ('jump' in options && options.jump));
+		const emojis = this._determineEmojis(
+			[],
+			!('stop' in options) || ('stop' in options && options.stop),
+			!('jump' in options) || ('jump' in options && options.jump),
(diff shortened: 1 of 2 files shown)


src/lib/util/RichDisplay.js
```diff
@@ -157,7 +157,14 @@ class RichDisplay {
 	 * @private
 	 */
 	_determineEmojis(emojis, stop) {
-		if (this.pages.length > 1 || this.infoPage) emojis.push(this.emojis.first, this.emojis.back, this.emojis.forward, this.emojis.last, this.emojis.jump);
+		if (this.pages.length > 1 || this.infoPage) {
+			for (const prop in this.emojis) {
```
bdistin: Objects do not maintain order. This will lead to the reactions being somewhat randomized.

src/lib/util/RichDisplay.js
```diff
@@ -91,11 +91,11 @@ class RichDisplay {
 
 	/**
 	 * Sets emojis to a new set of emojis
-	 * @param {RichDisplayEmojisObject} emojis An object containing replacement emojis to use instead
+	 * @param {RichDisplayEmojisObject} emojis An object containing replacement emojis to use instead with the option to remove options that are not wanted
 	 * @returns {RichDisplay} this RichDisplay
 	 */
 	setEmojis(emojis) {
-		Object.assign(this.emojis, emojis);
+		this.emojis = emojis;
```
bdistin: The old code should stay here. Instead, make the jump an option exactly as stop is an option now.

PR #44, opened 2017-10-04
Update help.js

Skillz4Killz:
Better appearance on Mobile

Still having complaints from users that now the text is too clumped and would look better if there was a space after each one? Any ideas how to do that. Ive tried doing \n but no luck


bdistin:
Awaiting fix, if non is provided by Sunday, this pr will be closed.

kyranet:
Handled by #57 by @pedall now. No changes has been made in the time requested, this PR is closed now.

PR #56, opened 2017-10-14
Fix RichMenu empty page and typings

HellPie:
There is always an empty page added due to the inclusion of just added options to the count of missing options to add, plus the typings for a RichMenu option suggested "description" instead of "body".


bdistin:
There are several other instances of description -> body s that need to be fixed

HellPie:
Weird. I thought that was the only object defining RichMenu options.

bdistin:
See like: https://github.com/dirigeants/klasa/blob/master/tutorials/names.json#L42 It creates a section at the bottom that says Further Reading. With the appropriate links. The child page then would only include information on how it extends the parent/what's the same, whats different/and examples without repeating a lot of what's already written/the same in the parent class.

bdistin:
still waiting on the line 41 fix

PR #57, opened 2017-10-14
configurable menu time for RichDisplay

pedall:
- makes menu closing time configurable for developers
- fixes  help message code block to be viewed correctly on mobile
- added `jump` and `firstLast` to typedefs


src/lib/util/ReactionHandler.js
```diff
@@ -62,6 +62,12 @@ class ReactionHandler extends ReactionCollector {
 		this.prompt = this.options.prompt || 'Which page would you like to jump to?';
 
 		/**
+		 * The time until the reaction collector closes automatically
+		 * @type {number}
+		 */
+		this.time = this.options.time || 30000;
```
kyranet: This should be

```js
this.time === typeof this.options.time === 'number' ? this.options.time : 30000;
```

src/commands/System/help.js
```diff
@@ -27,6 +27,7 @@ module.exports = class extends Command {
 		const helpMessage = [];
 		for (let cat = 0; cat < categories.length; cat++) {
 			helpMessage.push(`**${categories[cat]} Commands**: \`\`\`asciidoc`);
+			helpMessage.push('');
```
kyranet: Instead of calling the push method twice, you can push two values into the same array in a single call.

bdistin:
@kyranet I believe this pr effects typings. Can you confirm?

kyranet:
It does, I can patch it in a while once I finish something important, give me 2 hours or so to make sure.

PR #58, opened 2017-10-16
Update CommandMessage.js

Cinegez:
Fix bug reported by Hutch, using multiple spaces makes Komada and Klasa drop the "[...]" part of the usage.

### DIFF
--- a/src/lib/structures/CommandMessage.js
+++ b/src/lib/structures/CommandMessage.js
@@ -172,7 +172,7 @@ class CommandMessage {
 	 */
 	static getArgs(cmdMsg) {
 		// eslint-disable-next-line newline-per-chained-call
-		const args = cmdMsg.msg.content.slice(cmdMsg.prefixLength).trim().split(' ').slice(1).join(' ').trim().split(cmdMsg.cmd.usageDelim !== '' ? cmdMsg.cmd.usageDelim : undefined);
+		const args = cmdMsg.msg.content.slice(cmdMsg.prefixLength).trim().split(/\s+/g).slice(1).join(' ').trim().split(cmdMsg.cmd.usageDelim !== '' ? cmdMsg.cmd.usageDelim : undefined);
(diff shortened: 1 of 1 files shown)


tech6hutch:
Thanks for fixing (or attempting to fix) this

src/lib/structures/CommandMessage.js
```diff
@@ -172,7 +172,7 @@ class CommandMessage {
 	 */
 	static getArgs(cmdMsg) {
 		// eslint-disable-next-line newline-per-chained-call
-		const args = cmdMsg.msg.content.slice(cmdMsg.prefixLength).trim().split(' ').slice(1).join(' ').trim().split(cmdMsg.cmd.usageDelim !== '' ? cmdMsg.cmd.usageDelim : undefined);
+		const args = cmdMsg.msg.content.slice(cmdMsg.prefixLength).trim().split(/\s+/g).slice(1).join(' ').trim().split(cmdMsg.cmd.usageDelim !== '' ? cmdMsg.cmd.usageDelim : undefined);
```
tech6hutch: Does the first split remove the command name? I see the second split splits on usageDelim.

kyranet: This has a serious flag: it'll work perfectly when you use space as usage delim, but it will not when the usageDelim is different, causing weird issues. Same goes for quoted string arguments.

kyranet:
Unfortunately, "the fix" for this is complicated. To be honest, Klasa (and Komada) should take account on the length of parameters and iterate over them by skipping the empty parameters until it reaches valid arguments (non-empty arguments).

I have asked @AoDude if he could patch this (the fix for this is beyond me, usage is my weakness), but told me it wasn't a bug, but the previous behauviour.

I also suggested the "infinite repeator tag" which captures everything, but it was a bad idea and could make the code much more complicated.

For now, I'll put Bug: Invalid, Help Wanted and Enhancement flags.

PR #62, opened 2017-10-24
Fixed some lazytypings in KlasaConsole.

kyranet:


kyranet:
Ported to #66 so this can be merged in Master instead.

PR #74, opened 2017-10-28
msg.send() updated

MrJacz:
Updated this as msg.send with embeds is borked as you couldn't do `msg.send({embed: lol});`


kyranet:
Honestly, I see no point of that, these two parameters are defaulted in `Message#sendMessage`.

https://github.com/dirigeants/klasa/blob/f67605d2a4adcb6c6efd80898b499d1aa1f23d47/src/extendables/sendMessage.js#L9

PR #80, opened 2017-11-09
toLowerCase the command name

Pandraghon:
Because of https://github.com/dirigeants/klasa/blob/b6df5b58161d9335412801559ec3c3af1594a159/src/monitors/commandHandler.js#L30
lower


bdistin:
```javascript
this.name = options.name || file[file.length - 1].slice(0, -3).toLowerCase();
```
options.name and any element of options.aliases being anything but lowercase is framework user (dev) error. Willing to concede in letting filenames not be all lowercase, but that also leads to the possibility of some unexpected supercedence when you have two commands test.js and Test.js, as depending on which one loads last, either or might be loaded.

Additionally, double check the creating commands tutorial, and ensure there is a note saying that command names and aliases must be lowercase

tech6hutch:
Not only does the creating commands tutorial not mention that names must be lowercase, but the example command it gives has a name with uppercase (`'yourCommandName'`).

If this is supposed to be user error, it should throw if you try to name a command with an uppercase letter, imo. It's unintuitive otherwise.

bdistin:
Wasting processing asserting that the dev correctly configures commands to work with the included commandHandler monitor (not even what someone else might prefer), is just as wasteful as correcting the input to the bias of our commandHandler monitor.

It should be mentioned in the tutorial though. (lowercase no spaces)

Pandraghon:
So... just : 
```js
this.name = options.name || file[file.length - 1].slice(0, -3).toLowerCase();
```

The VSC extension should have a note about that too

tech6hutch:
I meant checking up front, when commands are loaded (not run), if the name contains any capital letters. I can't imagine that would take up much processing power.

Not a big deal, of course; it would just help reinforce the requirement.

PR #81, opened 2017-11-10
Command handler improvements

bdistin:
startsWith is slower than regex.test with a ^, and especially so when cached, instead of recreated every time a message is run (also eliminates regex.exec to get the prefix length).


kyranet:
And RegExp#test is slower than this:

```javascript
startsWithPrefix(prefix, content) {
	for (let i = 0; i < prefix.length; i++)
		if (prefix[i] !== content[i]) return false;
	return true;
};
```

bdistin:
Under limited conditions only. While a map cached regexp is consistently ~10m ops no matter the length. the for function goes from ~50m ops at prefix `!` to ~5m ops at prefix `ok google,`; Additionally it can be fooled into taking longer to return false by messages that start with what the prefix starts with like the following command and natural message `ok butterbot, flapjack time` and `ok but that idea sucks`.

*The above op references include one tested command message, and one tested non-command message*

That doesn't even include the other added benefits of not having to create RegExp instances, or execing them every time a command is run which actually lowered the ops to ~147k ops

PR #82, opened 2017-11-10
Better fr-FR

Pandraghon:


bdistin:
Can you explain the (e). Please note I don't speak, read, or write french.

Pandraghon:
In frensh, nouns have gender and the adjectives are accorded in both gender and number.
Ex; 
**a** required command => **une** commande requis**e**
**a** required event => **un** événement requis

bdistin:
maybe I asked the question poorly... most of your changes is from ...é to ...é(e)

Pandraghon:
It's a common practice in french. When you don't know if the noun will be 'male' or 'female' (singular or plural), you put the variants in `()`.
Ex: 
COMMAND_RELOAD_ALL: (type) => `✅ Tou(te)s les ${type} ont été rechargé(e)s.`,
If `type` is `commands` ("female"), it should display "Toutes les commands ont été rechargées."
If it's `events` ("male"), it should display "Tous les events ont été rechargés."

bdistin:
ah, ok, that explains it to me. Thank you

tech6hutch:
I had an idea. Could you (Pandraghon) just add logic to output the correct word based on what type was passed? Or would that be too much for a core language piece, Ao?

bdistin:
no, it can be a more complex arrow function... It's kinda there to do whatever you need to do for the language...

Pandraghon:
I don't think that's the best piece name translation, but ¯\\\_(ツ)\_/¯

tech6hutch:
So every type of piece is feminine except for commands in French? Interesting.

Pandraghon:
It depends on what word you choose for the translation
A finalizer could be translated as "une limite", which is feminine

kyranet:
Is this PR tested and working perfectly? (To be honest I don't know French)

Pandraghon:
The french-relative part should work correctly (for what I've tested).
I'm just unsure about the toTitleCase update. It works but is it the right thing to do ?

PR #84, opened 2017-11-15
better stopwatch

bdistin:


src/lib/util/Stopwatch.js
```diff
@@ -57,12 +59,57 @@ class Stopwatch {
 	}
 
 	/**
+	 * If the stopwatch is running or not
+	 * @since 0.4.0
+	 * @readonly
+	 * @type {boolean}
+	 */
+	get running() {
+		return Boolean(this._end);
```
kyranet: It could be `return this._end === null;`, since it's either null or a number.

tech6hutch:
The typings weren't updated

MrJacz:
Give me a sec

PR #90, opened 2017-11-21
Fix incorrect KlasaClientConfig typing

KenanY:
ClientOptions passed to discord.js should be under the `"clientOptions"`
property of KlasaClientConfig, rather than intersected with
KlasaClientConfig itself. This is what the code seems to expect, so this
patch updates the typings accordingly.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -953,6 +953,7 @@ declare module 'klasa' {
 	}
 
 	export type KlasaClientConfig = {
+		clientOptions?: ClientOptions;
 		prefix?: string;
 		permissionLevels?: PermissionLevels;
 		clientBaseDir?: string;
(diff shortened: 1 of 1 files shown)


UnseenFaith:
Might actually be the case in release, or that it will be the case. It makes sense to pass them as one due to the fact that Klasa is actually technically a Discord.js client. This is still up for debate though.

kyranet:
Yeah, but it's much easier to just pass the D.JS options with a property of `clientOptions` being type of `DiscordJS.ClientOptions`, as well as it can avoid some conflicts and helps organizing the code. Nonetheless, it is true that it breaks the consistency with some pieces (for example in Komada, you have both `exports.help` and `exports.conf`, but all properties are merged in a single object for Klasa).

However, the current state is that D.JS options are passed via `options.clientOptions` (althought it could be renamed to `discordjsOptions` or with another name, so we can "configure" other things such as console and providers.

PR #91, opened 2017-11-21
Fix `RichMenu#run()` return value type

KenanY:
`RichMenu#run()` returns the same type as `RichDisplay#run()`, which is
a `Promise<ReactionHandler>`.

PR #92, opened 2017-11-21
Swap `_determineEmojis()` signatures

KenanY:
From the code, it seems like `RichDisplay#_determineEmojis()` has the
four-argument signature, whereas `RichMenu#_determineEmojis()` has the
two-argument signature. The typings have them mixed up, so this patch
swaps the signatures to reflect the code.


KenanY:
Moving to https://github.com/dirigeants/klasa/pull/91.

PR #98, opened 2017-11-26
Add commandUnknown, and piece events

bdistin:


bdistin:
@kyranet Can you add typings for these new events. Also it appears there isn't typings for commandRun.

MrJacz:
Can I? 🤔

bdistin:
I guess, make a pr to this pr. Simpler that kyra can commit directly, but sure take a shot at it.

PR #99, opened 2017-11-26
Typings for new-events done

MrJacz:
commandRun event added as well as unknowCommand + all the piece events

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -115,11 +115,20 @@ declare module 'klasa' {
 		// Klasa Command Events
 		public on(event: 'commandError', listener: (msg: CommandMessage, command: Command, params: any[], error: Error) => void): this;
 		public on(event: 'commandInhibited', listener: (msg: CommandMessage, command: Command, response: string|Error) => void): this;
+		public on(event: 'commandRun', listener: (msg: CommandMessage, command: Command, params: any[], response: any) => void): this;
+		public on(event: 'commandUnknown', listener: (msg: DiscordMessage, command: string) => void): this;
 
 		// Klasa Console Custom Events
 		public on(event: 'log', listener: (data: any, type: string) => void): this;
 		public on(event: 'wtf', listener: (failure: Error) => void): this;
 		public on(event: 'verbose', listener: (data: any) => void): this;
+
+		// Klasa Piece Events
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -115,12 +115,21 @@ declare module 'klasa' {
 		// Klasa Command Events
 		public on(event: 'commandError', listener: (msg: CommandMessage, command: Command, params: any[], error: Error) => void): this;
 		public on(event: 'commandInhibited', listener: (msg: CommandMessage, command: Command, response: string|Error) => void): this;
+		public on(event: 'commandRun', listener: (msg: CommandMessage, command: Command, params: any[], response: any) => void): this;
+		public on(event: 'commandUnknown', listener: (msg: DiscordMessage, command: string) => void): this;
```
kyranet: `msg: ExtendedMessage`, it's not a message from Discord, but an extended one.

kyranet: `msg: ExtendedMessage`. It's not a DiscordMessage, but a much extended one.

typings/index.d.ts
```diff
@@ -115,12 +115,21 @@ declare module 'klasa' {
...
 
 		// Klasa Console Custom Events
 		public on(event: 'log', listener: (data: any, type: string) => void): this;
 		public on(event: 'wtf', listener: (failure: Error) => void): this;
 		public on(event: 'verbose', listener: (data: any) => void): this;
 
+		// Klasa Piece Events
+		public on(event: 'pieceLoaded', listener: (piece: Piece) => void): this;
```
kyranet: I'd suggest against the use of Piece since it's an abstract class and shall not be used directly, however, they get instances of each kind of piece.

typings/index.d.ts
MrJacz: Right (My bad)

typings/index.d.ts
MrJacz: Welp what to use then considering Piece is all pieces aka Commands, Events, Languages, Monitors etc

MrJacz: I guess I could do `Commnad|Monitor|Language|Event` etc

kyranet: Oh, that's right... keep that.

Issue #94, opened 2017-11-25
Suggestion: RegExp prefix's


MrJacz:
A way to be able to have custom Prefix with regex's aka `this.client.config.prefix` could be `new RegExp(/!/g)` for example

bdistin:
Uh, this was a thing... If it's broken, that's a bug.

MrJacz:
Confirmed bug

kyranet:
`this.client.config.prefix` has never been RegExp but `string`/`string[]`. This suggestion breaks several pieces including SettingGateway. Althought SGv2 will just throw a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError).

MrJacz:
I'm sure it was a thing in August I remember using it but I could be tripping this might break SGv2 but maybe there could be a RegexPrefix option in KlasaClientConfig so that RegexPrefix is the main prefix that couldn't be overridden aka `Hey Skyra, ping` would work always even if you used conf to change the prefix so it is not overridable.

kyranet:
Just like Komada 0.19.7's usage were so bugged the repeator tag was useless and it was capturing from the argument's position to the end (infinite catching), it got fixed in 0.20.0 with Ao's rewrite and that behaviour dissapeared. There has not been an infinite catching feature in Komada, nor in Klasa, although a bug allowed it to exist.

Same happens here with the prefixes, `this.client.config.prefix` being a RegExp (but guild prefixes defaulted to `string`|`string[]`). That was an unintended behaviour that worked due to a bug, but since the bug is now fixed, that does not longer work.

MrJacz:
I see

MrJacz:
@kyranet What is your opinion on the `regexPrefix` option in KlasaClientConfig? I think it would become handy

MrJacz:
as PrefixMention shouldn't be overidden and is only meant for mentioni prefix for the bot...

kyranet:
My opinion is that `regexPrefix` is a nice-to-have feature, but most developers will not find it handy. (Keep in mind that [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) is a feature that many newbie developers find hard to use.

MrJacz:
I can agree with it being a nice feature to have and very true that noobie developers dont know the power of Regular expresions

PR #103, opened 2017-11-29
regexPrefix KlasaClientOption for Defaulted non-overidden regex prefix

MrJacz:
this was suggested by me in #94 mainly because prefix KlasaClientOption doesn't support Regexp's because of SG, This is handy as it cannot be overridden meaning you can have `prefix` in KlasaClientConfig but that can be changed this cannot be changed which comes in handy :), P.S Comments on using `msg.content.match()` and not `regexp.exec()`


MrJacz:
Ugh ofc
it does this closing and fixing this ughhhhh

PR #121, opened 2017-12-10
ClientStorage

kyranet:
### Description of the PR

This PR also includes a refactor for **Gateway**, removing **GatewaySQL** completely, and adds two new classes: `GatewayStorage`, which `Gateway` now extends, and `ClientStorage`, which extends `GatewayStorage` aswell. This change has been made as both Gateway and ClientStorage shared too many similarities and Gateway is robust enough to handle many kinds of behaviours.

`ClientStorage` is one of the up-coming features for 0.5.0-dev, using SettingGateway's engine, for global configuration, such as UserBlacklist and GuildBlacklist.

This PR has no direct impact in `0.5.0-dev` and it does not contain any breaking changes for the developer settings-wise, however, this **does** modify a lot of internals from the settings system.

EDIT: With bdistin's refactor, there's a breaking change, check the changelog below.

**Added**:
- Added `constants` and `Util.mergeDefault`
- Added `GatewayStorage`, containing `SettingGateway`'s core.
- Added a new gateway called `clientStorage`, for client-wide configs.

**Changed**:
- **[BREAKING]** Merged `this.client.config` to `this.client.options`. (This change has been made to not confuse `client.config` (`KlasaClientConfig` type) with `client.configs` (`Configuration` instance).
- Refactored `Gateway` to extend `GatewayStorage`.

**Removed**:
- Removed `GatewaySQL`.

### Semver Classification


src/lib/settings/GatewayDriver.js
```diff
@@ -63,29 +62,29 @@ class GatewayDriver {
 		return {
```
UnseenFaith: This could probably be moved as well, but that would require it to be a function and I'm not sure that's the best idea in constants.

src/lib/settings/ClientStorage.js
```diff
@@ -0,0 +1,227 @@
...
+		 */
+		this.data = null;
+	}
+
+	/**
+	 * The data schema Klasa uses for client configs.
+	 * @since 0.5.0
+	 * @readonly
+	 */
+	get defaultSchema() {
```
UnseenFaith: This could be moved though.

src/lib/util/util.js
```diff
@@ -94,7 +95,27 @@ class Util {
...
 
 	/**
-	 * Applies an interface to a class|
+	 * Sets default properties on an object that aren't already specified.
+	 * @param {Object} def Default properties
+	 * @param {Object} given Object to assign defaults to
+	 * @returns {Object}
+	 * @private
+	 */
+	static mergeDefault(def, given) {
```
UnseenFaith: Not really much reason that this code should be duplicated no? This is equivalent to the one Discord.js uses and it should probably be used instead of duplicating it specifically for Klasa, or at the very least:

```js
const { Util: { mergeDefault } } = require("discord.js"):

class Util {
  static mergeDefault (...args) {
    return mergeDefault(...args); // or Util.mergeDefault to dispute naming issues
  } 
}
```

src/lib/settings/GatewayStorage.js
```diff
@@ -0,0 +1,233 @@
...
+	 */
+	static _parseSQLValue(value, schemaPiece) {
+		if (typeof value !== 'undefined') {
+			if (schemaPiece.array) {
+				if (value === null) return schemaPiece.default.slice(0);
+				if (typeof value === 'string') value = tryParse(value);
+				if (Array.isArray(value)) value = value.map(val => GatewayStorage._parseSQLValue(val, schemaPiece));
+				return value;
+			}
+			if (schemaPiece.type === 'any') {
```
UnseenFaith: Switchcase would probably be better here instead of if else if else if

src/lib/util/util.js
bdistin: I tried that originally, but then it fails on this.mergeDefault in d.js...

UnseenFaith: We talked about this on the server and determined It was probably best left alone to keep destructuring from breaking.

src/lib/settings/GatewayDriver.js
kyranet: They're not constants but variables, since they depend on `client`. That's why I decided to put there. Nonetheless this can be changed later if we find a better way.

src/lib/settings/GatewayStorage.js
kyranet: Oh yeah, could be. Next commit comes with a cleanup for this method 👍

PR #124, opened 2017-12-12
Put guildsSchema and clientStorageSchema into constants.js

Pandraghon:
### Description of the PR
Put gateways' schema alongside the resolver in `constants.js`.

typings are not up-to-date, 'cause I don't know how to type.

- Removed: constants.GATEWAY_RESOLVERS.GUILDS, constants.GATEWAY_RESOLVERS.USERS, constants.GATEWAY_RESOLVERS.CLIENT_STORAGE, \<GatewayDriver>.guildsSchema, \<GatewayDriver>.clientStorageSchema

- Added: constants.GATEWAYS.GUILDS, constants.GATEWAYS.USERS, constants.GATEWAYS.CLIENT_STORAGE

### Semver Classification


src/lib/util/constants.js
```diff
@@ -45,38 +45,112 @@ exports.DEFAULTS = {
...
+				if (value) return value;
+			}
 
-		throw new Error('The parameter <Guild> expects either a Guild ID or a Guild Instance.');
+			throw new Error('The parameter <Guild> expects either a Guild ID or a Guild Instance.');
+		},
+		schema: {
+			prefix: {
+				type: 'string',
+				default: this.client.options.prefix,
```
kyranet: The `this` keyword has no context here, therefore doubtfully it'll work. That's why I kept them in GatewayDriver. Resolvers are fine because of this line:

https://github.com/dirigeants/klasa/blob/531b938b0388a5719510428730d321da059ad17b/src/lib/settings/GatewayDriver.js#L160

bdistin:
There was issues with default schemas not being a getter as they are now, so this pr added more complexity than solved iirc. (+1 month stale)

PR #130, opened 2017-12-15
Make all piece defaults available for customization in KlasaClientOptions

kyranet:
### Description of the PR

This PR aims to provide ~~an option to default the value `Command.deletable` when not providen.~~ developers a new option in `KlasaClientOptions` where they can insert their default values for each kind of piece. For example:

```javascript
const { Client } = require('klasa');

new Client({
    prefix: '$',
    pieceDefaults: {
        commands: { quotedStringSupport: true },
        extendables: { enabled: false },
        monitors: { ignoreOthers: false }
    }
}).login('...');
```

**Added**:
- Added a new option in `KlasaClientOptions` to allow developers to set their own defaults for each kind of piece.
- ~~Added `KlasaClientOptions.deletableCommands` to change the default value for `Command.deletable` when not given.~~

**Changed**:
- Updated the documentation and some outdated typings.

**Removed**:
- Removed `KlasaClientOptions.quotedStringSupport` in favor of the new per-piece defaults.

**Notes**:
- ~~I have noticed that `Command.quotedStringSupport` will always be `true` even when the developer wants it to be disabled (`false`), and this change will do the same for `Command.deletable` (perhaps use `Object.assign` to modify the defaults in the line 44 of `src/lib/structures/Command.js`?)~~

### Semver Classification

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,4 @@ docs
 
 \.vscode/
 \.idea/
+desktop.ini
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ NOTE: For the contributors, you add new entries to this document following this 
 
 ### Added
 
-- [[#130](https://github.com/dirigeants/klasa/pull/130)] Added `KlasaClientOptions.cmdDeleting` to change the default value for `Command.deletable` when not given. (kyranet)
+- [[#130](https://github.com/dirigeants/klasa/pull/130)] Added a new option in `KlasaClientOptions` to allow developers to set their own defaults for each kind of piece. (bdistin)
 - [[#128](https://github.com/dirigeants/klasa/pull/128)] Added the `monitorError` event. (kyranet)
 - [[#125](https://github.com/dirigeants/klasa/pull/125)] Added the default gateways to `GatewayDriver` defaulted by null to reflect in the documentation. (kyranet)
 - [[#125](https://github.com/dirigeants/klasa/pull/125)] Added `Schema` (the previous got renamed to `SchemaFolder`), reducing duplicated code and bringing more code consistency. (kyranet)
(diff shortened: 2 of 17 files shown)


UnseenFaith:
The deletable addition is useless, it's already added into Klasa via constants

kyranet:
Not really, we're also doing that here:

https://github.com/dirigeants/klasa/blob/d590ffd11da4957e679fadaf8c68d14fc1ec7a0d/src/lib/structures/Command.js#L177

And was bdistin who implemented that line. Thing is, constants are constants, not variables, non-default KlasaClientOptions don't affect the original options, therefore it doesn't override and we need to default them manually.

UnseenFaith:
I really see no need for this.. if you want a command to be deletable it should be explicitly set, just like if you wanted a permission level not to be default you would set it yourself.

kyranet:
By that same logic we have two options:

- 1: Remove `KlasaClientOptions.quotedStringSupport`, as the developers should explicitely set it, whilst also increasing code duplication.
- 2: Modify the constants after `KlasaClientOptions`'s merge.
- 3: `Object.assign();` `KlasaClientOptions.quotedStringSupport` with the Command's constants and default them correctly (if it's explicitely set to false but the option for KlasaClient is set to true, it'll be false).
- 4: We also add `KlasaClientOptions.cmdDeleting`, following the same logic as `quotedStringSupport`.

Nonetheless, you still have the choice to set `cmdDeleting` to `false` in your KlasaClientOptions and set all commands' `deletable` to `true` if you want, this is not a breaking change, it just helps developers reducing their CommandOptions a little bit by defaulting properties that are the same in all of them.

UnseenFaith:
It's already defaulted to false though for commands. The only reason you would ever want to use this is if you wanted to enable it for all commands, and that seems very unlikely. And there's a big difference when it comes to configuring a command and Code Duplication. Commands should be configured exactly how you want them, not ambiguously by outside sources.

UnseenFaith:
QuotedStringSupport is also a special cookie because it's left out of the defaulting for Commands, now whether that was intended or not is a different story though.

PR #145, opened 2017-12-27
Fixed ClientStorage configs not updating across shards

kyranet:
### Description of the PR

This PR adds a missing check in SchemaFolder's cross-shard updating (to not update itself) and fixes the event `configUpdateEntry` to also update ClientStorage's entries.

This PR also fixes a *SchemaFolder's cross-shard updater* bug: the first parameter, `path`, accepts an array, and a stringified array has the format of `management,roles,administrator` instead of `["management","roles","administrator"]`, resulting on errors for sharded bots.

**Fixes**:

- Fixed ClientStorage configs not updating across shards, and added a missing check in some `broadcastEval`s.

### Semver Classification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,7 @@ NOTE: For the contributors, you add new entries to this document following this 
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


bdistin:
I thought ClientStorage internally updated cross shard?

kyranet:
No, it wasn't. That's why I made this PR. The bit from #150 was some residual code from an old attempt

PR #148, opened 2017-12-27
Fixed the reaction handler using endpoints without permissions

### LINKED ISSUE #147: Bug: RichDisplays in DMs Destroy Logs
### Describe the issue

Having a RichDisplay sent in a DM will cause havoc on your logs. The commands work perfectly. The rich display works perfectly. The pages change perfectly. It does everything that it should do but it keeps sending this error to your log for every reaction that clicked. So if a user clicks 2-3 times and you have thousands of users on a popular command it destroys logs. 

There should be a way to catch/ignore this error somehow because as I said the command works perfectly. The users have no issues. But the developer logs are insanely spammed.

```javascript
Uncaught Promise Error: 
DiscordAPIError: Cannot execute action on a DM channel
 at item.request.gen.end (node_modules/discord.js/src/rest/handlers/RequestHandler.js:79:65)
at then (node_modules/snekfetch/src/index.js:218:21)
 at <anonymous>
 at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

```
1) Create a rich display command to run in a DM
2) Use reactions
3) Check logs
```

### Expected an actual behaviour

Either a workaround or some sort of method to avoid this log is necessary. The only issue here is with the error log and not the RichDisplay or command. All i really need is a way to ignore this error from my console.

### Further details

- **discord.js version**: latest
- **node.js version**: 9+
- **Klasa version**: latest
- [x] I have tested the issue on latest master. Commit hash:

kyranet:
### Description of the PR

Fixes #147 

When using RichDisplay/RichMenu in channels where the bot does not have the `MANAGE_MESSAGES` permission or in DMs, the ReactionHandler instance was trying to delete the reactions from the user, and clear all reactions when the ReactionHandler stops/finishes. Both actions can only be executed in a guild and with the previously mentioned permissions.

- Fixed ReactionHandler trying to remove reactions from other users or clear all reactions from a message even when the client doesn't have permissions in the guild or is in DMs

### Semver Classification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,7 @@ NOTE: For the contributors, you add new entries to this document following this 
 
 ### Fixed
 
+- [[#148](https://github.com/dirigeants/klasa/pull/148)] Fixed ReactionHandler trying to remove reactions from other users or clear all reactions from a message even when the client doesn't have permissions in the guild or is in DMs [#147](https://github.com/dirigeants/klasa/issues/147). (kyranet)
 - [[#142](https://github.com/dirigeants/klasa/pull/142)] Fixed a critical bug in nested objects when using the JSON provider. Note: `Object.assign` doesn't merge nested objects. (kyranet)
 - [[#141](https://github.com/dirigeants/klasa/pull/141)] Fixed `KlasaConsoleConfigs` not defaulting correctly. (bdistin)
 - [[#141](https://github.com/dirigeants/klasa/pull/141)] Fixed wrong sharding behaviour when using PM2 in a non-sharded bot. (bdistin)
--- a/src/lib/util/ReactionHandler.js
+++ b/src/lib/util/ReactionHandler.js
@@ -126,7 +126,7 @@ class ReactionHandler extends ReactionCollector {
 	 * @readonly
 	 */
 	get unreactable() {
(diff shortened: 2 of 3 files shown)


bdistin:
#147 is invalid because of #149 and vice versa. Making this pr invalid aswell. RichDisplays require manage messages.

kyranet:
It's not because #147 nor #149, this PR makes RichDisplay able to run in DMs or without permissions, EMBED_LINKS is still required, but MANAGE_MESSAGES is too much priviledge in many servers, and Discord doesn't give them in DMs yet

Skillz4Killz:
Just to clarify. 

#147 does NOT cause #149 . This fix only helped reveal 149 more to me as I didnt notice it before with the console bothering me so much. In reality even without this PR, the bot still requires clicking twice to change the page.

This PR helps clean up the issue where a developers logs are not spammed. The issue with #149 Is a separate issue altogether which I believe is being caused because RichDisplay does not change the page when a reaction is removed by a user. When the bot can not remove a reaction automatically, the user now has to click twice to remove and then add again.

In my opinion, you guys should add a listener to accept the removal of a reaction which will also turn the page IF the bot doesnt have the permissions to remove the reaction automatically.

PR #152, opened 2018-01-02
Per-gateway configurable options (users, guilds and clientStorage)

kyranet:
### Description of the PR

At the current status of Klasa, the developer is unable to configure each built-in gateway (`users`, `guilds` and `clientStorage`) separately. This PR implements this as a feature, and to do so, users will be able to configure every gateway.

This PR will be more useful in the future, specially if the `SettingOptions` type expands.

**Added**:

- The possibility to configure each default gateway separately.

**Changed**:

- **[BREAKING]** Changed `KlasaClientOptions.provider.engine` to `KlasaClientOptions.providers.default`.

### Semver Classification

### DIFF
--- a/src/lib/settings/GatewayDriver.js
+++ b/src/lib/settings/GatewayDriver.js
@@ -180,7 +180,8 @@ class GatewayDriver {
 		if (!this.client.methods.util.isObject(schema)) throw 'Schema must be a valid object or left undefined for an empty object.';
 
 		options.provider = this._checkProvider(options.provider || this.client.options.providers.default);
-		if (options.provider.cache) throw `The provider ${options.provider.name} is designed for caching, not persistent data. Please try again with another.`;
+		const provider = this.client.providers.get(options.provider);
+		if (provider.cache) throw `The provider ${provider.name} is designed for caching, not persistent data. Please try again with another.`;
 		options.cache = this._checkProvider('collection');
 
 		const gateway = new Gateway(this, name, validateFunction, schema, options);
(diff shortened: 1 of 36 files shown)


src/lib/settings/GatewayDriver.js
```diff
@@ -179,7 +179,7 @@ class GatewayDriver {
 		validateFunction = validateFunction.bind(this);
 		if (!this.client.methods.util.isObject(schema)) throw 'Schema must be a valid object or left undefined for an empty object.';
 
-		options.provider = this._checkProvider(options.provider || this.client.options.provider.engine || 'json');
+		options.provider = this._checkProvider(options.provider || this.client.options.gateways.defaultProvider.engine || 'json');
```
bdistin: defaultProvider is a string, and there is no need to double default it. (... || 'json')

src/lib/settings/GatewayStorage.js
```diff
@@ -37,7 +37,7 @@ class GatewayStorage {
 		 * @type {string}
 		 * @readonly
 		 */
-		Object.defineProperty(this, 'providerName', { value: provider || this.client.options.provider.engine || 'json' });
+		Object.defineProperty(this, 'providerName', { value: provider || this.client.options.gateways.defaultProvider || 'json' });
```
bdistin: There is no reason to double default defaultProvider (... || 'json')

kyranet:
Yes, I agree with that way of passing connection and database options via `KlasaClientOptions.providers`. Reverted that change.

PR #153, opened 2018-01-06
fix blacklist not ignoring owner

QuantumlyTangled:
### Description of the PR

This PR just fixes the blacklist inhibitor not ignoring the bot owner.

-

### Semver Classification

### DIFF
--- a/src/inhibitors/blacklist.js
+++ b/src/inhibitors/blacklist.js
@@ -3,7 +3,7 @@ const { Inhibitor } = require('klasa');
 module.exports = class extends Inhibitor {
 
 	async run(msg) {
-		if (msg.author === this.client.owner) return;
+		if (msg.author.id === this.client.options.ownerID) return;
 		if (this.client.configs.userBlacklist.includes(msg.author.id)) throw true;
 	}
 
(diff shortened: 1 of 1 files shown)


src/inhibitors/blacklist.js
```diff
@@ -3,6 +3,7 @@ const { Inhibitor } = require('klasa');
 module.exports = class extends Inhibitor {
 
 	async run(msg) {
+		if (msg.author === this.client.owner) return;
```
kyranet: For performance, I'd suggest `this.client.options.ownerID`, as `this.client.owner` is a getter:

https://github.com/dirigeants/klasa/blob/bb46f404a40448c880f0c0d4662a894f27dc7977/src/lib/Client.js#L296-L298

Soumil07:
What if the owner hasn't passed ownerID in his clientOptions?

MrJacz:
https://github.com/dirigeants/klasa/blob/master/src/lib/Client.js#L408

Soumil07:
Oh. I thought it did that in the getter.

QuantumlyTangled:
In that case, you might also want to change other inhibitors.

QuantumlyTangled:
Or I can.

bdistin:
In node 9 a get is as fast as a property access. And a reference compare has always been faster than a string compare. The way this pr was originally done (msg.author === this.client.owner) is the correct way to perform the check.

PR #182, opened 2018-02-12
Optimizations

bdistin:
### Description of the PR
Eliminates storing full paths to all pieces, and to stores over and over. Replaces functionality with getters that build the paths (which only cost once per class), and the core boolean property.

Adds a shortcut to store from the piece; so now in commands to call another command you can use this.store.get(commandOrAlias). (also used internally)

Extracts Tag.pattern so it isn't redefined over and over every time usage is parsed.

Simplifies store.load, so you don't have to pass the dir anymore. You just pass the relative file path from the store folder to the piece (usually just the filename) and a boolean if core or not. Pretty much just private use here.

**Changed**:

- **[BREAKING]** Refactored all pieces and stores to extend Piece and Store respectively.

> TypeScript users are the ones affected by this change, JavaScript users using [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) with options (format used in the docs) should not have issues.

**Fixed**:

- Reduced the amount of data stored per-piece, resulting on a RAM usage decrease.
- Fixed `MessageOptions` being not correctly applied to the *send* aliases.

### Semver Classification


bdistin:
I probably wrecked typings, although it should make them easier/less duplication.

PR #199, opened 2018-02-23
Command string-return sending

tech6hutch:
### Description of the PR
Most commands send a message at the end. This PR allows you to return a string (or a string[]) from a command to send it to the relevant channel (or edit the previous message in that channel, if `cmdEditing` is turned on). This works with subcommands, as well.

I changed a few core commands to take advantage of this new feature. However, it may be better to leave the core commands using msg.sendMessage explicitly, so they'd still work if the user disables command string-return sending.

Originally, I modified commandHandler.js directly, but then I decided that making it a finalizer might be better. Users can then easily disable this feature by transferring the finalizer and setting `enabled: false`.

- A new finalizer, commandSendReturn.js, which sends the return value from commands, if it's a string or array of strings
- (Possibly) a rewrite of the core commands to take advantage of this feature

### Semver Classification

### DIFF
--- a/src/commands/Admin/conf.js
+++ b/src/commands/Admin/conf.js
@@ -26,27 +26,27 @@ module.exports = class extends Command {
 
 	get(msg, [key]) {
 		const { path } = this.client.gateways.guilds.getPath(key, { avoidUnconfigurable: true, piece: true });
-		return msg.sendMessage(msg.language.get('COMMAND_CONF_GET', path.path, path.resolveString(msg)));
+		return msg.language.get('COMMAND_CONF_GET', path.path, path.resolveString(msg));
 	}
 
 	async set(msg, [key, ...valueToSet]) {
 		const { path } = await msg.guild.configs.update(key, valueToSet.join(' '), msg.guild, { avoidUnconfigurable: true, action: 'add' });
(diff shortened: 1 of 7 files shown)


kyranet:
The idea of this PR is nice, provides consistency with `throw` (which makes the command exit as a non-successful run and reply the user with whatever the reason is). However, it's quite limited in terms of options, and expanding them may be more expensive.

tech6hutch:
I'll see if I can expand it, without making it too involved to run on every command.

bdistin:
This breaks the consistency that Messages should be returned from Commands; Something that Finalizers and the commandSuccess events inherently expect to get. If this is even used once, this prevents any finalizer from it's previous capability of depending on the sent message.

That's before considering that this only works for content, and only creates extra overhead for commands that you need to send that isn't exclusively content.

Not to mention this is going too far with hand holding.

tech6hutch:
Can this be reopened please? I've changed things, and I think you'll be more pleased with it. It addresses the expected Message return, provides more flexibility to have more than just string content, removes all overhead for the check other than a single object property access, and is opt-in, so users can't rely on hand-holding too much.

PR #205, opened 2018-02-25
Command return sending (improved)

tech6hutch:
### Description of the PR
Most commands send a message at the end. This PR allows you to return a string (or any other type that `KlasaMessage#sendMessage` supports) from a command to send it to the relevant channel (or edit the previous message in that channel, if `cmdEditing` is turned on).

I changed a few core commands to take advantage of this new feature, as examples.

#### Details and Benefits
- To send the return value, all you have to do is add `subcommand: true` to your Command options.
- Supports all arguments and features of `sendMessage`. Specify `sendReturn: 'apply'` if you'd like to return an array of args for `sendMessage`. (`true` or `'call'` assume you're returning one argument.)
  - This means you can send, e.g., a JS codeblock by returning `['const code = here()', { code: 'js' }]`, with `sendReturn: 'apply'`.
- Before the `subcommand` option was added, you could send strings returned from the "subcommands" with `msg.send(this[firstArg](restOfArgs))` in the `run` method. The `subcommand` option is useful, but when using it, you must explicitly send messages in every method. Not anymore, as command-return-sending works especially well with subcommands.
* Reduces code duplication, and reduces complexity in commands that nest a lot of things inside a call to `sendMessage` (retrieving language strings, etc.).
* Provides feature parity with throwing, since you can already `throw` strings to send them. Throw for an error message, return for a success message.
* Finalizers are passed the Message object gotten from sending the return value, not the return value itself, keeping consistency.

#### Changes From Previous Pull Request
- Finalizers still get KlasaMessage objects now, even if you use the `sendReturn` feature
- Much more flexible in what you can send and what options you send with, as all of `KlasaMessage#sendMessage`'s features are supported, while still allowing simple string returns
- Simplifies the check for whether to send a command's return value to a simple object property access (plus a string-to-boolean conversion if it _is_ a command whose return value should be sent), so it's more efficient, especially for commands that don't use this feature
- Fully backwards compatible*, so no matter what an existing command happens to return (e.g., if they happen to return something other than a KlasaMessage, against the current specs), it won't be accidentally sent as a message
\* Unless a user has made their own `sendReturn` option, but this risk is, of course, negligible.

- A new Command option, `sendReturn`
- An edit to commandHandler.js, which sends the return value from commands if the command's `sendReturn` option is truthy
- A rewrite of the core commands to take advantage of this feature

### To-do (If This PR is Accepted)
- Update docs and typings
- Change the rest of the core commands

### Semver Classification


tech6hutch:
(I'll be away for most of the evening, so I won't be able to do any requested changes right away.)

bdistin:
While this does address some of the bigger problems I brought up. (like self-destructing responses finalizers) It still doesn't address the biggest, and that's the whole concept of this pr is too much hand-holding.

Not only that, but this updated pr adds additional state and overhead to commands, to avoid using msg.send() inside the command itself. While you are attempting to bring parity to `throw`, you misunderstand that to handle both error objects (unexpected or otherwise), and friendly responses; that the friendly response can't be an object, like Message is.

If you want to add this functionality to your bot, you of course are free to and you should do it in your user space. This however doesn't belong in the core framework.

tech6hutch:
@bdistin okay, I understand. Thank you for the response. Would something like this be appropriate for the Pieces repo, in some form?

bdistin:
There is no reason why a "LordOfCommandHandlers" couldn't be made for the pieces-repo, so long as it passes kyra's review.

tech6hutch:
Alright, cool. I might do that. Not sure how I'll handle the <Command>.sendReturn default, though.

PR #226, opened 2018-03-09
Added CreatingPointsSystems.md to the guides

kyranet:
### Description of the PR

The **social module**, or **points based systems** is a feature that is common in many bots, and it was very requested in our userbase. This PR adds a guide everyone can follow and understand better how can they do it with SettingGateway, taking advantage of our built-in per-user configs.

- Added CreatingPointsSystems.md to the guides

### Semver Classification


guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+		this.client.gateways.users.schema.add('experience', {
+			type: 'integer',
+			default: 0,
+			configurable: false
+		});
+	}
+}
+```
+
+Where we'd check if the schema has the key `experience`. If it doesn't, we add it as a new key, with type `integer` (doubtfully we'll use `float` on this) and make it unconfigurable for the built-in userconf command so the end users do not cheat by modifying their stats.
```
Tylertron1998: "Where we'd check if the schema..." should be changed to "this is where we'd check if the schema..."

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+
+> **Note**: The name of the key can be any, can be `xp`, `points`... but we will use this one for this guide.
```
Tylertron1998: "The name of they key can be any..." should be changed to "The name of they key can be anything...", and also "but we will use this one for this guide." uses "this" twice; suggest "but we will use this one for the guide."

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Setting the monitor
+
+Now that we have set up the schema, we will want to create a monitor
```
Tylertron1998: missing a colon at the end (I think?)

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+
+};
+
+```
+
+Alternatively, we can create the `init` method and ensure the users' schema always has our key.
+
+## Level up!
+
+Some social bots have level up messages, but, how do we set it up? There are two ways to achieve this:
```
Tylertron1998: Possibly get rid of the comma after "but"? Reading it out-loud, sounds rather odd, the pause in between "but" and "how do we..."

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+
+1. We calculate the current level and the next level on the fly. This system is, however, harder to implement and it processes a lot of maths, but it's also RAM friendly for massive bots. We won't cover this in this guide.
```
Tylertron1998: Again, double this; "We won't cover this in this guide". Maybe change to "We won't cover this in the guide".

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+		// Calculate the next value for experience.
+		const nextValue = msg.author.configs.experience + 1;
+
+		// Cache the current level.
+		const currLevel = msg.author.configs.level;
+
+		// Calculate the next level.
+		const nextLevel = Math.floor(0.1 * Math.sqrt(nextValue + 1));
+
+		// Update the user's configuration entry by adding 1 to it, and update the level aswell.
```
Tylertron1998: Aswell isn't a word, "also" would be the recommended one to use.

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+		await msg.author.configs.update(['experience', 'level'], [nextValue, nextLevel]);
+
+		// If the current level and the next level do not equal, then it has increased, and you can send the message.
```
Tylertron1998: "If the current level and the next level do not equal" might be better to change to "If the current level and the next level are not the same"

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+			await msg.send(`Congratulations! You leveled up to level **${currLevel}**!`);
+		}
+	}
+
+	// Init
+
+}
+```
+
+Optionally, you can check if `nextLevel === msg.author.configs.level` is true and update a single key instead, but the speed difference is negligible and since [SettingGateway v2.1](https://github.com/dirigeants/klasa/pull/179), the key `level` will not be updated if it did not change. As well as this overload is much faster than the JSON object overload, previously used as the only way to update multiple values.
```
Tylertron1998: "As well as this overload is much faster than..." would be better worded as: "As well, this overload is much faster".

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Creating Our Commands
+
+To allow users know their current amount of points and level, we will create two commands:
```
Tylertron1998: "To allow user know..." => "To allow users to know..."

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+
+Let's create a file in `commands/Social/points.js` with the following contents:
+
+```javascript
+const { Command } = require('klasa');
+
+module.exports = class extends Command {
+
+	constructor(...args) {
+		super(...args, { description: 'Check your amount of points' });
```
Tylertron1998: The description may be better as "Check how many points you have."

guides/Other Subjects/CreatingPointsSystems.md
```diff
@@ -0,0 +1,161 @@
...
+
+Let's create a file in `commands/Social/level.js` with the following contents:
+
+```javascript
+const { Command } = require('klasa');
+
+module.exports = class extends Command {
+
+	constructor(...args) {
+		super(...args, { description: 'Check your amount of points' });
```
Tylertron1998: Same here, the description may be better as "Check how many points you have.".

PR #236, opened 2018-03-14
Broadcast eval functions

bdistin:
### Description of the PR
Makes use of broadcastEval functions over strings which should allow somewhat simpler code (in this case a bit safer), and the ability to lint those portions of the code to ensure code-quality and accuracy. 

However none of these changes are tested, and will need to be tested by someone using traditional sharding.

-

### Semver Classification


bdistin:
D.JS 's function does not work as expected.

PR #238, opened 2018-03-16
Minor usage refactor for readability and (probably) performance

tech6hutch:
### Description of the PR
Minor usage refactors for readability and (probably) performance
Also fixes a usage min/max bug

- Small refactors
- Doc fixes
- Change Possible.resolveLimit to non-static

### Semver Classification
PATCH

### DIFF
--- a/src/lib/usage/Possible.js
+++ b/src/lib/usage/Possible.js
@@ -54,7 +54,7 @@ class Possible {
 	 * Resolves a limit
 	 * @since 0.2.1
 	 * @param {string} limit The limit to evaluate
-	 * @param {"min"|"max"} limitType The type of limit
+	 * @param {string} limitType The type of limit
 	 * @returns {number}
 	 * @private
 	 */
--- a/src/lib/usage/Possible.js
+++ b/src/lib/usage/Possible.js
@@ -1,4 +1,7 @@
+// The types towards the beginning require integer min/max limits
 const minMaxTypes = ['str', 'string', 'int', 'integer', 'num', 'number', 'float'];
+// The index of the last type in minMaxTypes that requires integer min/max
+minMaxTypes.intLimitLastIndex = 3;
 const regexTypes = ['reg', 'regex', 'regexp'];
 
 /**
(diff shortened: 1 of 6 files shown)


tech6hutch:
Maybe hold off on merging this, for now. I think the usage system could use more commentary, so I I'll add that, as I go through it.

src/lib/usage/Tag.js
```diff
@@ -9,7 +9,7 @@ class Tag {
 	 * @since 0.2.1
 	 * @param {string} members The tag contents to parse
 	 * @param {number} count The position of the tag in the usage string
-	 * @param {boolean} required The type of tag (required or optional)
+	 * @param {number} required The type of tag (0 for optional, anything else for required)
```
bdistin: 0 optional, 1 semi-required, 2 required

typings/index.d.ts
```diff
@@ -965,11 +965,11 @@ declare module 'klasa' {
 		public max: number;
 		public regex: RegExp;
 
-		private static resolveLimit(limit: string, type: string): number;
+		private resolveLimit(limit: string, limitType: "min" | "max"): number;
```
tech6hutch: Please let me know if this is not a valid TS type

MrJacz: just `string`

tech6hutch:
I see, will fix

bdistin:
No, just no to the last commit

src/lib/usage/Possible.js
```diff
@@ -1,4 +1,7 @@
+// The types towards the beginning require integer min/max limits
 const minMaxTypes = ['str', 'string', 'int', 'integer', 'num', 'number', 'float'];
+// The index of the last type in minMaxTypes that requires integer min/max
+minMaxTypes.intLimitLastIndex = 3;
```
MrJacz: Shouldn't this just be a variable instead of adding a property to an array?

tech6hutch:
But if it's hardcoded, it'll break if any new types are ever added. I guess a comment to update the number when adding new types would suffice

bdistin:
The Possible class is exactly as it should be in master at this time. This pr should not change how it is from master.

src/lib/usage/Possible.js
tech6hutch: That would also work. I put it on the array since it was related to it.

tech6hutch:
I'm done with this PR now, unless there's anything else.

PR #261, opened 2018-04-05
Update Getting Started to add Discord.JS

FayneAldan:
### Description of the PR

Added Discord.JS to the `npm install` command on the Getting Started page

N/A

### Semver Classification

### DIFF
--- a/guides/Getting Started/GettingStarted.md
+++ b/guides/Getting Started/GettingStarted.md
@@ -5,13 +5,13 @@ Time to take the plunge! Klasa is on NPM and can be easily installed.
 > I assume you know how to open a command prompt in a folder where you want to install this. Please don't prove me wrong.
 
 ```sh
-npm install --save klasa
+npm install --save discordjs/discord.js klasa
 ```
 
 optionally if you want to use the bleeding edge development version (not guaranteed to be stable):
 
 ```sh
-npm install --save dirigeants/klasa
+npm install --save discordjs/discord.js dirigeants/klasa
 ```
 
 ### Using Klasa


Tylertron1998:
"About:
Klasa is an OOP discord.js bot framework which aims to be the most feature complete, while feeling like a consistant extension of discord.js.

Originally based on Komada, Klasa has become a ship of Theseus and so much more. You can only vaguely see similarities between the two frameworks anymore..." If you don't know you are meant to install discord.js, then you are an idiot.

guides/Getting Started/GettingStarted.md
```diff
@@ -5,13 +5,13 @@ Time to take the plunge! Klasa is on NPM and can be easily installed.
 > I assume you know how to open a command prompt in a folder where you want to install this. Please don't prove me wrong.
 
 ```sh
-npm install --save klasa
+npm install --save discordjs/discord.js klasa
```
Tylertron1998: This is wrong, stable uses d.js stable.

FayneAldan:
It was already discussed in the Discord server that master Discord.js should be used

Tylertron1998:
So it seems. From the package.json, it appears that stable uses master d.js.

My apologise. 

My point still stands, however, you should KNOW that you need to install d.js. Should we tell people they need to install node too? And have a system to run it on? All of these things you should know, if not, you probably shouldn't be developing a bot.

FayneAldan:
Yes, you SHOULD know that Discord.JS is needed, but nothing ever said that **master** Discord.JS is needed.

tech6hutch:
"My point still stands, however, you should KNOW that you need to install d.js."
Not necessarily. It's a peer dep, but if it was a normal dep, it would be installed automatically.

PR #264, opened 2018-04-06
Fixed optional type in typescript

ShayBox:

-

### Semver Classification

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1300,7 +1300,7 @@ declare module 'klasa' {
 	}
 
 	export class Type {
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -1300,7 +1300,7 @@ declare module 'klasa' {
 	}
 
 	export class Type {
-		public constructor(value: any, parent: ?type);
+		public constructor(value: any, parent: type?);
```
kyranet: It's actually 

```ts
public constructor(value: any, parent?: type);
```

PR #265, opened 2018-04-07
KlasaConsole typings fix

tech6hutch:
### Description of the PR
Fixed the KlasaConsole TS error

- Changed KlasaConsole to not implement Console

### Semver Classification

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1152,7 +1152,7 @@ declare module 'klasa' {
 		private static _parse(pattern: string): number;
 	}
 
-	export class KlasaConsole implements Console {
+	export class KlasaConsole {
 		private constructor(client: KlasaClient, options: KlasaConsoleConfig);
 		public readonly client: KlasaClient;
 		public readonly stdout: NodeJS.WritableStream;


tech6hutch:
I still get errors.
```
node_modules/klasa/typings/index.d.ts(1285,19): error TS8020: JSDoc types can only be used inside documentation comments.
node_modules/klasa/typings/index.d.ts(1286,22): error TS2314: Generic type 'Map<K, V>' requires 2 type argument(s).
node_modules/klasa/typings/index.d.ts(1287,24): error TS2314: Generic type 'Map<K, V>' requires 2 type argument(s).
node_modules/klasa/typings/index.d.ts(1301,31): error TS2314: Generic type 'Map<K, V>' requires 2 type argument(s).
The terminal process terminated with exit code: 1
```

Should these also be fixed, and if so, how?

tech6hutch:
@bdistin please let me know if the types I specified for the maps in your Type class are wrong.

typings/index.d.ts
```diff
@@ -1305,9 +1282,9 @@ declare module 'klasa' {
 		public value: any;
 		public is: string;
 
-		private parent: ?Type;
-		private childKeys: Map;
-		private childValues: Map;
+		private parent: Type | null;
```
bdistin: the ? makes the type nullable, although it was in the wrong place, it should be `private parent?: Type;`

tech6hutch: Ah okay. I would have thought a `?` on the property name would mean the property itself might not exist.

tech6hutch:
Okay, that last commit seems to fix Extendable#extend. It can now be either a function or a getter.

PR #268, opened 2018-04-10
Cache user responses in TextPrompt#responses to have access to the messages

pedall:
### Description of the PR
New Collection to cache user input messages in TextPrompt#responses

-

### Semver Classification

### DIFF
--- a/src/lib/usage/TextPrompt.js
+++ b/src/lib/usage/TextPrompt.js
@@ -125,7 +125,6 @@ class TextPrompt {
 		 * @type {Tag}
 		 * @private
 		 */
-
 		this._currentUsage = {};
 
 		/**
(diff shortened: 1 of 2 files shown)


typings/index.d.ts
```diff
@@ -996,6 +996,7 @@ declare module 'klasa' {
 		public promptTime: number;
 		public promptLimit: number;
 		public quotedStringSupport: boolean;
+		public responses: Collection;
```
bdistin: reasonably sure this needs <string, KlasaMessage>, but I defer to @kyranet

kyranet: Exactly, since `Collection<K, V> extends Map<K, V>`, it's a generic, and must have both values specified. In this case it'd be:

```ts
public responses: Collection<string, KlasaMessage>;
```

pedall: done

src/lib/usage/TextPrompt.js
```diff
@@ -124,7 +125,15 @@ class TextPrompt {
 		 * @type {Tag}
 		 * @private
 		 */
+
```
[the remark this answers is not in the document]
pedall: done

PR #274, opened 2018-04-18
[Awaiting Stability] ESM Support

bdistin:
### Description of the PR
Adds EMCAscript Modules support to Klasa, so that you can write your bots with .mjs: import/export syntax, or as you normally would now.

This breaks require.main.filename, and I don't have a solution for that currently, except hardcoding with the src/ folder, as the vs code extension suggests. A better solution will be required before merging.

This will not merge until the feature is no-longer behind the `--experimental-modules` flag, but will track changes with the master branch for anyone who wants to give this a try.

-

### Semver Classification

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,7 +6,7 @@ resources:
       endpoint: dirigeants
 
 jobs:
-  - template: lint.yml@templates
+  #- template: lint.yml@templates
 
   - template: docs.yml@templates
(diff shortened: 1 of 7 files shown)


kyranet:
Added the `Status: Blocked` and `Meta: V8` labels as this PR is pending for a newer V8 implementation of `import()` (currently experimental in Node.js 10, V8 6.6).

PR #275, opened 2018-04-19
Fix typings

tech6hutch:
### Description of the PR
Allowed commands to return `null` (which commands in core do, already). I also updated the finalizer run method to note it can receive `null` for `mes`.

Also fixed a couple other things.

-

### Semver Classification


tech6hutch:
This PR is done

tech6hutch:
Oh right, I think we had discussed that. Sorry.

PR #285, opened 2018-04-26
Remove duplicate identifier (Typings)

FayneAldan:
### Description of the PR
Fixes an error in Typings that prevented TypeScript from compiling.

### Semver Classification


kyranet:
That's actually fixed in #275, it meant usersSchema, not guildsSchema.

PR #293, opened 2018-05-03
Update PieceStores.md

Soumil07:
### Description of the PR
Updates PieceStores to include an example on a working rawEvent store.

-

### Semver Classification


Soumil07:
I'll resolve the travis errors tmrw.

KingDGrizzle:
This documentation could be better improved, and would probably be better off rewritten from scratch for the raw store part.
Plus, theres still pieces from the old example for music. Just saying

guides/Other Subjects/PieceStores.md
```diff
@@ -33,122 +33,50 @@ class Something extends Piece {
...
 
-		// Otherwise wrap the next song
-		return this.wrapLink(next.id);
-	}
+	constructor(client, store, file, core, options = {}) {
+		super(client, store, file, core, options);
 
-	wrapLink(id) {
-		return `https://youtu.be/${id}`;
+		// not a lot of options, but im keeping it simple for the sake of the guide
```
KingDGrizzle: I mean this comment could be improved to be like `// Enable this raw event by default if not specifically disabled in the options`

guides/Other Subjects/PieceStores.md
```diff
@@ -33,122 +33,50 @@ class Something extends Piece {
...
 	}
 
 }
 
-module.exports = Genre;
+module.exports = RawEvent;
 ```
 
-Here we see some heavy customization. Such as if you need to define special enable/disable behavior. But Genre extending Piece makes our piece very easy to actually make now:
+Not a lot of options here, so most times you won't be needing your constructor anyway.
```
KingDGrizzle: Imo, `There aren't a lot of options here, so you won't need the constructor most of the times`, or something similar

guides/Other Subjects/PieceStores.md
```diff
@@ -158,38 +86,56 @@ This is great and all, but we need to register these pieces/store:
...
-		this.genres = new GenreStore();
+		// Make a new RawEventStore and attach it to client
+		this.rawEvents = new RawEventStore();
 
-		// Register the GenreStore to be loaded, init, and available
+		// Register the RawEventStore to be loaded, init, and available
 		// to be used as an arg to be looked up in commands
-		this.registerStore(this.genres);
-
-		// Note: You will likely want to make a custom Argument for this new piece type
```
KingDGrizzle: This comment should probs remain.

guides/Other Subjects/PieceStores.md
```diff
@@ -158,38 +86,56 @@ This is great and all, but we need to register these pieces/store:
...
-		.getNext(this);
-
-	// nextSong should now be a pseudo random song based on the
-	// genre seeds and what has recently played.
+const { Event } = require('klasa');
+const { Constants: { Events } } = require('discord.js');
+
+module.exports = class extends Event {
+
+	async run(data) {
```
KingDGrizzle: Here, it should probably be packet, so be consistent with d.js and to represent what it actually it :)

guides/Other Subjects/PieceStores.md
```diff
@@ -158,38 +86,56 @@ This is great and all, but we need to register these pieces/store:
...
 	}
 
 }
 
-new MySwankyMusicBot().login('token-goes-here');
+new MyBot().login('token-goes-here');
 ```
 
-Then we use it in our player class I completely made up earlier like so:
+Now to actually make our raw events work! This code goes in events/raw.
```
KingDGrizzle: `Now, to make our raw event store actually work, we're going to add this code in events/raw to run on every raw event received`

guides/Other Subjects/PieceStores.md
```diff
@@ -158,38 +86,56 @@ This is great and all, but we need to register these pieces/store:
...
+		if (this.client.rawEvents.has(event)) {
+			const rawEvent = this.client.rawEvents.get(event);
+			if (rawEvent.enabled) rawEvent.run(data.d);
+		}
+	}
+
+};
+```
+
+Now an example raw event, we'll have messageReactionAdd, which will go in rawEvents/messageReactionAdd
```
KingDGrizzle: `Now, for an example event, we'll have messageReactionAdd, which will go in rawEvents/messageRreactionAdd and will run on all added reactions` or close to this.

guides/Other Subjects/PieceStores.md
```diff
@@ -158,38 +86,56 @@ This is great and all, but we need to register these pieces/store:
...
+
+```javascript
+const RawEvent = require('../RawEvent');
+
+module.exports = class extends RawEvent {
+
+	/* eslint-disable camelcase */
+	async run({ channel_id, message_id, user_id }) {
```
KingDGrizzle: Probably a bad idea to deconstruct this data. Maybe just show it with `run (data)` and console.logging it, since data packets are weird af and can have any value, and ye + no more random eslint comment.

Tylertron1998: Why disable it, and why not use camelCase?

bdistin: it's raw, it comes from discord as snek_case

Tylertron1998: Hm, that does make a little more sense, I suppose.

Soumil07:
I'm on vacation until the 15th. I will update to requested changes by the 18th.

bdistin:
closed because of #325

PR #299, opened 2018-05-08
fix: Escape strings in Util#initClean

KingDGrizzle:
### Description of the PR
Strings like a bot token can have the `.` character, which can cause the regex to misreplace. This fixes that by escaping any special characters from the password, email and token. Better safe than sorry!

- Escape all strings in initClean

### Semver Classification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,7 @@ NOTE: For the contributors, you add new entries to this document following this 
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


KingDGrizzle:
Hrm. Seems like the `.` character doesn't actually break, as I was expecting. Would probably still be a good idea nontheless to escape these, but, waiting for your input

kyranet:
@KingDGrizzle is this PR tested?

PR #302, opened 2018-05-11
Fix command reload

kyranet:
### Description of the PR

When `Piece#reload()` is called on a piece that does not load correctly (due to an error thrown from the loader or failing to parse the code), it was calling `.init()` even when piece is `null` (which happens when the load is unsuccessful).

- Better error handling in `Piece#reload()`

### Semver Classification


bdistin:
This breaks the reload command... It's intended when you call reload() and the file doesn't exist, for the promise to be rejected.

PR #306, opened 2018-05-12
Added QueryBuilder, exterminated all SQL from SettingGateway

### LINKED ISSUE #287: Proposal Design: QueryBuilder
### Describe the proposal

The PR #228 went into stale due to the lack of an objective design and lack of a technical specification for QueryBuilder. This issue will remain opened until the design for QueryBuilder gets fully written, specified **and** merged into the master branch.

I'm opening this issue as QueryBuilder is part of the v1.0.0 RoadMap and therefore, its design should start taking priority so it can be written, tested and merged before discord.js 12 does, that way we can land the long-awaited Klasa 1.0.0 with discord.js 12 release.

The current (and last written) version of QueryBuilder looked like this:

```javascript
// Create a QueryBuilder for PostgreSQL
this.qb = new QueryBuilder(this.client, {
	// Declare the boolean type, with the datatype name as "BOOL"
	// and defaults false and true
	BOOLEAN: { name: 'BOOL', default: [false, true] },
	// Declare the float type, defaults to 0
	FLOAT: { name: 'DOUBLE PRECISION', default: 0 },
	// Declare the integer type, defaults to 0
	INTEGER: { name: 'INTEGER', default: 0 },
	// Declare the text type, defaults to NULL
	TEXT: { name: 'TEXT' },
	// Declare the varchar type, defaults to NULL, the property of
	// size tells QB to use VARCHAR(MAX_LENGTH)
	VARCHAR: { name: 'VARCHAR', size: true }
}, {
	// When using arrays, PG converts INTEGER to INTEGER[]. So we want
	// to append '[]' at the end of the type
	arrayWrap: (type) => `${type}[]`
});
```

However, this design has not been approved by the team. I would also like to open this to everyone who wants to contribute in the design of QueryBuilder (perhaps we can take a look to Sequelize and Knex, and get the best of them for some aspects?).

### LINKED ISSUE #311: [1.0.0 Roadmap] Remove all SQL state from the library (and schema).
The library should not care anything about what the provider is using in the background. This also allows for each sql provider to be more customized to that individual database, rather than trying to find a set of sql syntax/types common to all sql databases. As well as reduction of core code (bloat) and finally massive simplification of SettingGateway.

Potentially add a SQL QueryBuilder Util with overridable defaults, to aid in reduction of duplicate code/making SQL Providers more maintainable.

kyranet:
### Description of the PR

Fixes #311 
Fixes #287 

As requested for a long time, **ALL** SQL state has been removed without track from the entire SettingGateway code. **ALL** of it has been moved to providers, and **ALL** providers shall update to support the new code.

**Added**:

- Added `Util.isPrimitive`.
- Added `constants.DEFAULTS.QUERY_BUILDER`.
- Added `SQLProvider#updateColumn`.
- Added `SQLProvider#parseUpdateInput`, `Provider#parseUpdateInput`.
- Added `QueryBuilder` class util.

**Changed**:

- Changed `SQLProvider#createTable`. SettingGateway will not provide columns, for consistency with JSON providers. Instead, retrieve the columns from `Gateway`.
- Changed `SQLProvider#addColumn` to have the arguments `table: string, columns: SchemaFolder | SchemaPiece`.
- Made `Gateway`'s constructor public in typings.
- Made `Configuration#_existsInDB` nullable to denote Configuration instances that have never sync.
- Tweaked initial synchronization for performance.

**Removed**:

- Removed `GatewayGuildResolvable` type from typings.
- Removed `JSONProvider#set` and `JSONProvider#insert`.
- Removed `Provider#sql` and `SQLProvider#sql`.
- Removed `SchemaPiece#_generateSQLDatatype` and `SchemaPiece._parseSQLValue`.
- Removed `SchemaPiece#sql`.
- Removed `SchemaFolderAddOptions.sql`.
- Removed `GatewayStorage#sqlSchema`, `SchemaFolder#sqlSchema`, and `SchemaPiece#sqlSchema`.
- Removed `Role` and `GuildMember` overloads for `GuildResolvable`.

### Semver Classification


Soumil07:
And they said Vlad was the code breaker...

src/lib/util/QueryBuilder.js
```diff
@@ -0,0 +1,120 @@
...
+		 * @private
+		 */
+		this.formatDatatype = formatDatatype || ((name, datatype, def = null) => `${name} ${datatype}${def !== null ? ` NOT NULL DEFAULT ${def}` : ''}`);
+	}
+
+	/**
+	 * Get a datatype
+	 * @since 0.5.0
+	 * @param {string} type The datatype to get
+	 * @returns {QueryBuilderDatatype}
```
bdistin: `?QueryBuilderDatatype`

src/lib/util/QueryBuilder.js
```diff
@@ -0,0 +1,120 @@
...
+	 * @property {Function} [formatDatatype] The datatype formatter for the SQL database
+	 * @property {Function} [arrayResolver] The specialized resolver for array keys
+	 */
+
+	/**
+	 * @since 0.5.0
+	 * @param {Object<QueryBuilderDatatype>} datatypes The datatype to insert
+	 * @param {QueryBuilderOptions} [options = {}] The default options for all datatypes plus formatDatatype
+	 */
+	constructor(datatypes = {}, { array = () => 'TEXT', resolver = null, type = null, arrayResolver, formatDatatype } = {}) {
```
bdistin: `{ array = () => 'TEXT', resolver = null, type = null, arrayResolver, formatDatatype }` should be defaulted in constants and mergeDefault, instead of here.

Also, consider that all options in both datatypes and options are optional. This really should all be one parameter, options, mergeDefault, then `const { array, resolver, type, arrayResolver, formatDatatype, ...datatypes } = options;`

src/lib/util/QueryBuilder.js
```diff
@@ -0,0 +1,120 @@
...
+		if (!isObject(datatypes)) throw `Expected 'datatypes' to be an object literal, got ${new Type(datatypes)}`;
+
+		// Default the options for QueryBuilderDatatype
+		mergeDefault(DATATYPES, datatypes);
+
+		// Merge defaults on all keys
+		for (const key of Object.keys(datatypes)) {
```
bdistin: shouldn't this be Object.entries() since you are getting the value the very next line?

src/lib/util/QueryBuilder.js
```diff
@@ -0,0 +1,120 @@
...
+			const value = datatypes[key];
+			if (!isObject(value)) throw `Expected 'datatypes.${key}' to be an object literal, got ${new Type(value)}`;
+			mergeDefault({ array, resolver, type }, value);
```
bdistin: I am not sure I understand merging references of array, resolver, and type into the value. It seems like these should be qb state, since they are unchanging to this instance. I could be wrong about this comment.

src/lib/util/util.js
```diff
@@ -210,14 +210,24 @@ class Util {
...
 	}
 
+	/**
+	 * Check whether a value is a primitive
+	 * @since 0.5.0
+	 * @param {*} value The value to check
+	 * @returns {boolean}
+	 */
+	static isPrimitive(value) {
+		return ['string', 'bigint', 'number', 'boolean'].includes(typeof value);
```
bdistin: Should cache this array either as module scoped or as a static property.

src/lib/util/QueryBuilder.js
```diff
@@ -12,27 +12,24 @@ class QueryBuilder {
...
 	 */
 
 	/**
 	 * @since 0.5.0
-	 * @param {Object<QueryBuilderDatatype>} datatypes The datatype to insert
 	 * @param {QueryBuilderOptions} [options = {}] The default options for all datatypes plus formatDatatype
 	 */
-	constructor(datatypes = {}, { array = () => 'TEXT', resolver = null, type = null, arrayResolver, formatDatatype } = {}) {
+	constructor(options = {}) {
+		const { datatypes, array, resolver, type, arrayResolver, formatDatatype } = mergeDefault(options, QUERYBUILDER);
```
bdistin: No, I meant flatten datatypes into the rest, and use the rest operator to extract them.

Should be:
```js
const { array, resolver, type, arrayResolver, formatDatatype, ...datatypes } = mergeDefault(options, QUERYBUILDER);
```
ofcourse with the flattening done in constants.DEFAULTS.QUERYBUILDER aswell. For logical code seperation, can include a `// datatypes section` in the constants file.

PR #310, opened 2018-05-14
SG Auto-Resolving

UnseenFaith:
### Description of the PR
Adds the functionality for creating and resolving configurations automatically. This option is disabled by default, and can be enabled via SchemaPiece#resolve. Klasa will automatically resolve any and all resolvable values from the Schema (where possible).

Current implementation is merely a concept that needs to be built upon because the current code is somewhat ugly (SettingResolver in particular).

Syncing in SG should probably get a rewrite with this as well because it's currently horrendous. SG should automatically sync, and outside sources can fetch && sync at the same time if necessary. This can also add room for Configuration#synced to determine whether or not it's up to date with the database.

A lot of documentation also needs to be added still.

- Adds Configuation#resolve(data) => Takes SG data and turns into a fully resolved object

### Semver Classification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -227,6 +227,7 @@ NOTE: For the contributors, you add new entries to this document following this 
 
 ### Removed
 
+- [[#401](https://github.com/dirigeants/klasa/pull/401)] Removed `Settings#waitSync` in favor of `Settings#sync(?false);`. (kyranet)
 - [[#398](https://github.com/dirigeants/klasa/pull/398)] Removed the `Settings#update(keys: string[], values: any[])` overload. (kyranet)
 - [[#391](https://github.com/dirigeants/klasa/pull/391)] Removed `Util.getIdentifier()`. (kyranet)
 - [[#391](https://github.com/dirigeants/klasa/pull/391)] Removed Gateways' second caching layer. (kyranet)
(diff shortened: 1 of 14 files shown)


src/lib/parsers/SettingResolver.js
```diff
@@ -14,9 +14,10 @@ class SettingResolver extends Resolver {
 	 * @param {string} name The name of the key being resolved
 	 * @returns {string}
 	 */
-	async user(data, guild, name) {
+	async user(data, guild, name, { resolve, def }, silent) {
```
bdistin: Missing docs for resolve, def, and silent. Also, I assume this breaks the "all optional parameters go in an options object as the last parameter" consistency rule.

Same goes for the many instances of the same pattern below.

src/lib/extensions/KlasaGuild.js
```diff
@@ -35,7 +35,7 @@ module.exports = Structures.extend('Guild', Guild => {
 		 * @returns {Object}
 		 */
 		toJSON() {
-			return { ...super.toJSON(), configs: this.configs };
+			return { ...super.toJSON(), configs: this.configs.toJSON() };
```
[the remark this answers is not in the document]
UnseenFaith: It was more of to make it consistent with the other call, not necessarily to toJSON it, it can be removed if you want.

src/lib/parsers/SettingResolver.js
UnseenFaith: This will probably be change because I don't like this format at all.

PR #312, opened 2018-05-16
Abort the prompt if the user retypes the command

bdistin:
### Description of the PR
by checking if the content starts with the previously used prefix

-

### Semver Classification


Skillz4Killz:
This may end up breaking some valid prompts for developers using `.` prefix. As an example, suppose one of my arguments ask people to provide a numerical value and someone wants to provide .5

Example: the arg require is how much currency you wish to give. Someone wants to do 50 cents so they type .5 but the prefix is also `.` this would abort the prompt when the value is properly given.

Would it not be better to just do 
```js
if (message.command) abort the error
```

This solves the main problem of the user typing the command again during the prompt but it also goes beyond that and prevents users from typing a different command during a prompt.

Example: start a command and initiate the prompt for arg and then type `.help` which launches another command and the prompt continues.

bdistin:
This pr remains open as a reminder to come up with a solution that actually works in the TextPrompt rework.

PR #317, opened 2018-05-23
Remove double-quotes from en-US language

mcumbers:
### Description of the PR
Every other string uses single-quotes except these three...

-

### Semver Classification


bdistin:
That's because of [this](https://github.com/dirigeants/klasa-lint/blob/master/.eslintrc.json#L135) eslint rule where we allow avoidescape. So since that's ok by our style guide, what's existing is already perfectly fine.

I appreciate your contribution, but I don't think I will merge this particular pr. I hope this won't detour you from making contributions in the future though.

PR #318, opened 2018-05-23
fix(Loader): CommonJS modules appending duplicated modules on reload

kyranet:
### Description of the PR

When a CommonJS module loads another, it appends the `Module` instance inside `module.children`. But when you reload the same module, the old module is not replaced, but rather duplicated, causing a memory leak.

This PR fixes that issue by removing the old module (if it exists).

Thanks @cthuluhoop123 for pointing it out in https://github.com/AnIdiotsGuide/guidebot/pull/46

- Remove the old module from `Module#children`

### Semver Classification

### DIFF
--- a/src/lib/structures/base/Store.js
+++ b/src/lib/structures/base/Store.js
@@ -90,15 +90,15 @@ class Store extends Collection {
 		const loc = join(dir, ...file);
 		let piece = null;
 		try {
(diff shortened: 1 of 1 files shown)


bdistin:
fixed in bc8a37384e3575edc6ebd7d58c816d1eeab5d605

PR #322, opened 2018-05-26
Make Store.walk() resolve '.ts' files as well

jdenderplays:
### Description of the PR
Stores walk their dictionaries and load all '.js' files as pieces. This pull request make them load '.ts' files as well.

- Patch Store to load '.ts' files

### Semver Classification
Minor patch

### DIFF
--- a/src/lib/structures/base/Store.js
+++ b/src/lib/structures/base/Store.js
@@ -169,7 +169,7 @@ class Store extends Collection {
 	 */
 	static async walk(store, core = false) {
 		const dir = core ? store.coreDir : store.userDir;
-		const files = await fs.scan(dir, { filter: (stats, path) => stats.isFile() && extname(path) === '.js' }).catch(() => { fs.ensureDir(dir).catch(err => store.client.emit('error', err)); });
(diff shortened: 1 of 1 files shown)


MrJacz:
You compile TS to JS, you don't run typescript you compile it.

kyranet:
You cannot run `.ts` files in native Node.js, you need to compile them first.

jdenderplays:
This was intended to add support to various typescript hooks, not vanilla node. (i.e. ts-node or typescript-require)

ShayBox:
Should be implemented, very useful.
ts-node is nicer than having to compile.

ShayBox:
Why do you need a plugin for this, just merge the pr for people that use node registers for .ts extensions.

This doesn't impact non-ts users at all, non ts users wont have .ts files in their folders, and if you're worried about that, add a check to see if they're using a node register.

I'm tired of waiting a minute for my bot to compile every time I save.

kyranet:
1. The repository has been deleted, I can't reopen this PR, let alone merging it.
1. The check is flawed, it should check it's not a `.d.ts` file, which may be present in JS projects (like mine).
1. Registering `.ts` natively in klasa implies we need `ts-node` bundled as one of our dependencies.
1. This should be `Object.keys(require.extensions).filter(key => key !== '.json' && key !== '.node');`, then filter files that are in those extensions. This way, when `ts-node` is present, `.ts` will be added to `require.extensions`, therefore loaded.

PR #325, opened 2018-05-30
[RFC] docs: Raw Event store

KingDGrizzle:
### Description of the PR
Adds raw event store examples to the docs. Supersede  #293 

### Semver Classification


guides/Other Subjects/PieceStores.md
```diff
@@ -33,163 +33,110 @@ class Something extends Piece {
...
 
-		// Otherwise wrap the next song
-		return this.wrapLink(next.id);
-	}
+	constructor(client, store, file, core, options = {}) {
+		super(client, store, file, core, options);
 
-	wrapLink(id) {
-		return `https://youtu.be/${id}`;
+		// Enable this raw event by default if not specifically disabled in the options
```
bdistin: This is not the correct way to do this. You should show rawEventStore defaults in KlasaClientOptions instead

guides/Other Subjects/PieceStores.md
```diff
@@ -33,163 +33,110 @@ class Something extends Piece {
...
-		.getNext(this);
-
-	// nextSong should now be a pseudo random song based on the
-	// genre seeds and what has recently played.
+const { Event } = require('klasa');
+
+module.exports = class extends Event {
+
+	async run(packet) {
+		if (this.client.rawEvents.has(packet.t)) {
```
bdistin: has + get is an antipattern. Use get only, check truthy.

PR #340, opened 2018-06-13
Add support for extension injection

kyranet:
### Description of the PR

This PR implements a more abstracted solution to https://github.com/dirigeants/klasa/pull/322, which can support other libraries that inject extensions (let's say somebody does `ts-node` but for Coffescript!). This PR has no impact in the current system, instead, if an installed library injects an extension that can translate a language into valid JS, this will be "caugh" by the loader (if found) and loaded.

Assigning to @bdistin as this affects the loader.

- Added support for extension injection library

### Semver Classification

### DIFF
--- a/src/lib/structures/base/Store.js
+++ b/src/lib/structures/base/Store.js
@@ -170,7 +170,9 @@ class Store extends Collection {
 	 */
 	static async walk(store, core = false) {
 		const dir = core ? store.coreDir : store.userDir;
-		const files = await fs.scan(dir, { filter: (stats, path) => stats.isFile() && extname(path) === '.js' }).catch(() => { fs.ensureDir(dir).catch(err => store.client.emit('error', err)); });
+		const extensions = Object.keys(require.extensions).filter(key => key !== '.json' && key !== '.node');
+		const files = await fs.scan(dir, { filter: (stats, path) => stats.isFile() && extensions.includes(extname(path)) })
+			.catch(() => { fs.ensureDir(dir).catch(err => store.client.emit('error', err)); });
 		if (!files) return true;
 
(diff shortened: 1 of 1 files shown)


bdistin:
Almost certain this will not be compatible with ESM which is a planned part of the future of this package. So rather than get everyone used to this, and then complaining when it breaks because of ESM, I would rather not.

If people want to do custom loading, they should use a static extendable to overwrite Store.walk(). Even then, it will probably break with ESM, but atleast it's on the users to know that it may be changed, rather than the library saying new feature then months later removes feature for ESM.

ShayBox:
It's not like the ESM version of Klasa will be replacing every other version in the next year, not only is ESM behind a flag, it's also in the .mjs file extension, assuming node 11 ships with ESM out of the flag and not requiring the file extension node 10 will still be LTS for years to come, so Klasa will need a version that's maintained for anyone using LTS node.
`stable` branch is a version that supports LTS node
`master` branch is the version that can use latest node features

Regarding the actual topic of this pull: extension support, that's reasonable to not want a feature that would possibly not be implementable in a few years, so thank you for telling me how I can implement this in my bot: (replacing store.walk)

bdistin:
You misunderstand. Esm will always have .mjs filenames, and the esm branch supports both cjs and esm files. The only thing that makes it major pr breaking is it requires a higher version of node than 8.6 currently required. 10.4 will soon be required by d.js, so it's possible it will be downgraded to a minor version tag.

Dev-Yukine:
I think that Plugin support is way more important then Esm support because it offers your more advantages and i personal would choose it over esm at any time. its a shame that you wont implement it directly since Klasa at all is the most customizable Framework but this would offer you so much to advance this Framework like you want without hacky ways.

Dev-Yukine:
Oh my bad, i misread the titel.

PR #355, opened 2018-07-05
fix: Schedule#delete

kyranet:
### Description of the PR

This PR adds a small fix, a full fix would be checking whenever the client's config updated the schedules entry. (Pending for discussion about this)

- Fixed Schedule not deleting entries that do not exist in ClientStorage but are still cached in Schedule#tasks.

### Semver Classification

### DIFF
--- a/src/lib/schedule/Schedule.js
+++ b/src/lib/schedule/Schedule.js
@@ -162,11 +162,16 @@ class Schedule {
 	 * @returns {this}
 	 */
 	async delete(id) {
-		const _task = this._tasks.find(entry => entry.id === id);
-		if (!_task) throw new Error('This task does not exist.');
+		const task = this._tasks.find(entry => entry.id === id);
+		if (!task) {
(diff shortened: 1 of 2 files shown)


src/lib/schedule/Schedule.js
```diff
@@ -162,11 +162,16 @@ class Schedule {
 	 * @returns {this}
 	 */
 	async delete(id) {
-		const _task = this._tasks.find(entry => entry.id === id);
-		if (!_task) throw new Error('This task does not exist.');
+		const task = this._tasks.find(entry => entry.id === id);
+		if (!task) {
+			const possibleIndex = this.tasks.findIndex(entry => entry.id === id);
```
bdistin: duplicate logic, you need possibleIndex and splice regardless of `!task` or `task`

kyranet: Fixed

PR #356, opened 2018-07-05
fix: Document Timestamp#displayUTC and patch it up

KingDGrizzle:
### Description of the PR
Documents Timestamp#displayUTC in typings
And fixes the default parameter to actually be a Date instance

-

### Semver Classification

### DIFF
--- a/src/lib/util/Timestamp.js
+++ b/src/lib/util/Timestamp.js
@@ -46,7 +46,7 @@ class Timestamp {
 	 * @param {(Date|number|string)} [time=new Date()] The time to display in utc
 	 * @returns {string}
 	 */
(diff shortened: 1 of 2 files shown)


src/lib/util/Timestamp.js
```diff
@@ -46,7 +46,7 @@ class Timestamp {
 	 * @param {(Date|number|string)} [time=new Date()] The time to display in utc
 	 * @returns {string}
 	 */
-	displayUTC(time) {
+	displayUTC(time = new Date()) {
```
bdistin: Don't double default, it's already defaulted in Timestamp.utc()

PR #363, opened 2018-07-15
Add an option to reload all stores

TheFloppyBanana:
### Description of the PR
This creates a subcommand that reloads all stores.

- An `all` subcommand to reload all stores using the reload command

### Semver Classification

- [X] This PR adds methods or properties to the framework interface.


src/commands/Admin/reload.js
```diff
@@ -39,4 +40,18 @@ module.exports = class extends Command {
 		}
 	}
 
+	async all(message) {
+		const timer = new Stopwatch();
+		this.client.pieceStores.forEach(async (p) => {
```
kyranet:
```javascript
await this.client.pieceStores.map(store => store.loadAll().then(() => store.init()));
```

src/commands/Admin/reload.js
```diff
@@ -39,4 +40,18 @@ module.exports = class extends Command {
...
+		await Promise.all(this.client.pieceStores.map(async (p) => {
+			await p.loadAll();
+			await p.init();
+		}));
+		if (this.client.shard) {
+			await this.client.shard.broadcastEval(`
+				if (this.shard.id !== ${this.client.shard.id}) this.client.pieceStores.map(async (p) => await p.loadAll());
+			`);
+		}
+		return message.sendMessage(`Reloaded all stores. (Took: ${timer.stop().toString()})`);
```
bdistin: toString() is unnecessary in a template literal, it already implicitly calls toString();

src/commands/Admin/reload.js
```diff
@@ -39,4 +40,18 @@ module.exports = class extends Command {
...
 
+	async all(message) {
+		const timer = new Stopwatch();
+		await Promise.all(this.client.pieceStores.map(async (p) => {
+			await p.loadAll();
+			await p.init();
+		}));
+		if (this.client.shard) {
+			await this.client.shard.broadcastEval(`
+				if (this.shard.id !== ${this.client.shard.id}) this.client.pieceStores.map(async (p) => await p.loadAll());
```
bdistin: you also need to init all here...

src/commands/Admin/reload.js
TheFloppyBanana: Noted, will push a change about this.

bdistin:
also, instead of all, wouldn't it make more sense: `!reload everything`?

Tylertron1998:
Wouldn't it make more sense just to use a default subcommand - i.e. `!reload`; because you're reloading _everything_ anyway?

src/commands/Admin/reload.js
```diff
@@ -39,4 +40,21 @@ module.exports = class extends Command {
...
+		}));
+		if (this.client.shard) {
+			await this.client.shard.broadcastEval(`
+				if (this.shard.id !== ${this.client.shard.id}) this.client.pieceStores.map(async (store) => {
+					await store.loadAll();
+					await store.init();
+				});
+			`);
+		}
+		return message.sendMessage(`Reloaded all stores. (Took: ${timer.stop()})`);
```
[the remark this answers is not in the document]
TheFloppyBanana: Will do! :D

src/commands/Admin/reload.js
```diff
@@ -22,21 +23,39 @@ module.exports = class extends Command {
...
 		try {
 			const itm = await piece.reload();
+			const timer = new Stopwatch();
 			if (this.client.shard) {
 				await this.client.shard.broadcastEval(`
 					if (this.shard.id !== ${this.client.shard.id}) this.${piece.store}.get('${piece.name}').reload();
 				`);
 			}
-			return message.sendMessage(message.language.get('COMMAND_RELOAD', itm.type, itm.name));
+			return message.sendMessage(message.language.get('COMMAND_RELOAD', itm.type, itm.name, timer.stop.toString()));
```
bdistin: just `timer.stop()` we don't want the string of the function source.

src/commands/Admin/reload.js
```diff
@@ -22,21 +23,39 @@ module.exports = class extends Command {
 					if (this.shard.id !== ${this.client.shard.id}) this.${piece.name}.loadAll().then(() => this.${piece.name}.loadAll());
 				`);
 			}
-			return message.sendMessage(`${message.language.get('COMMAND_RELOAD_ALL', piece)} (Took: ${timer.stop()})`);
+			return message.sendMessage(`${message.language.get('COMMAND_RELOAD_ALL', piece, timer.stop())}`);
```
kyranet: Remove the template literal, it's unneded, since message.language.get returns a string.

TheFloppyBanana:
What if the keyword should be `all` and all changes should happen in the language file, for example:
`COMMAND_RELOAD_ALL` => `COMMAND_RELOAD_STORE`,
`COMMAND_RELOAD_EVERYTHING` => `COMMAND_RELOAD_ALL`

bdistin:
No, please do what we discussed last night. Keyword remains `everything` but is moved to the end, and is a default argument.

PR #366, opened 2018-07-19
Add a default Argument type

bdistin:
### Description of the PR
Say you want to do some subcommands, but you don't always want to require the user to define a subcommand to use. In comes the default argument.

Consider the usage `<option1|option2|option3|run:default>`. In this case, if none of the first three are matched the first argument will be the string `run`, making the run method called with subcommands enabled. But you are free to use the name however you want, making the default literal/subcommand method whatever you want.

-

### Semver Classification


src/arguments/default.js
```diff
@@ -0,0 +1,10 @@
+const { Argument } = require('klasa');
+
+module.exports = class extends Argument {
+
+	run(arg, possible, message) {
+		if (arg.toLowerCase() !== possible.name.toLowerCase()) message.args.splice(message.params.length, 0, undefined);
```
kyranet: You're lowercasing `possible.name` twice.

PR #379, opened 2018-07-31
Add: prefixCaseInsensitive client option

AdityaTD:
### Description of the PR
This PR adds a client option to toggle the prefix's case sensitivity.

-

### Semver Classification


bdistin:
There is a better way to implement this. When I get to a workstation, I will give more details

kyranet:
This should only edit this line:

https://github.com/dirigeants/klasa/blob/master/src/monitors/commandHandler.js#L75

And add `, 'i'` to allow insensitive-case check.

bdistin:
half of the right answer^

AdityaTD:
Oh I see what you mean, that's honestly a much efficient way.

PR #397, opened 2018-08-14
Fix timezone offset

KingDGrizzle:
### Description of the PR
Whenever you used Timestamp#displayUTC it'd remove Timestamp.timezoneOffset from the current date. The issue, is that, in timezones that are +NUMBER, the number would be negative, so instead of subtracting, it would add. (Thanks JS)

This PR fixes that inversion, by multiplying the offset by -1. This makes it to where someones whos on -NUMBER UTC gains hours, and those who are in +NUMBER UTC lose hours.

(Someone in the US help me test)

-

### Semver Classification

### DIFF
--- a/src/lib/util/Timestamp.js
+++ b/src/lib/util/Timestamp.js
@@ -153,7 +153,7 @@ class Timestamp {
 
 /* eslint-disable id-length */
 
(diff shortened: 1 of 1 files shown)


bdistin:
instead of double negative, change the - in Timestamp.utc() to +

PR #383, opened 2018-08-07
Schema Changes

UnseenFaith:
### Description of the PR
This PR aims to ease the use of Schema's throughout Klasa and it's plugins.

Something else profound goes here.

### Changes Proposed in this Pull Request (List new items in CHANGELOG.MD)

**tl;dr**:

- Rewrote SG's schemas entirely from scratch.
- Removes Schema.json files.
- Adds new Schema constructor.
- New Types for SG, and a way to add your own external types.
- More OOP approach to Schema's which will make it much easier to develop more features for SG without limitations.

---

**Added**:

- Added the `SETTING_GATEWAY_INVALID_FILTERED_VALUE` language key. (bdistin)
- Added `Base` class for schemas, extending `Map`. (Unseenfaith)
- Added `SchemaType` and `SchemaTypes` classes. (Unseenfaith)
- Added `SchemaPieceOptions.filter`. (kyranet)
- Added abstract method `SQLProvider#getColumns`. (kyranet)

**Changed**:

- **[BREAKING]** Changed SchemaFolder to extend Schema, which extends Map. All keys are now stored inside the map as opposed to being properties. (Unseenfaith)
- **[BREAKING]** Changed `Schema#add` and `Schema#remove` to be synchronous. They must be called before ready. (Unseenfaith)
- **[BREAKING]** Changed `GatewayDriver#register`'s arguments to move `schema`'s argument to `GatewayDriverRegisterOptions`. (kyranet)
- **[BREAKING]** Changed schemas' root to be a `Schema` instance instead of a `SchemaFolder` instance. (Unseenfaith)
- **[BREAKING]** Changed `GatewayStorage`'s schema argument to take a Schema instance. (kyranet)
- **[BREAKING]** Renamed events `configCreateEntry`, `configDeleteEntry`, and `configUpdateEntry` to `settingsCreateEntry`, `settingsDeleteEntry`, and `settingsUpdateEntry` respectively. (kyranet)
- **[SEMVER-MAJOR]** `SchemaFolder#add` will not longer throw on conflict, but edit the SchemaPiece or update the SchemaFolder. (kyranet)
- `SchemaPiece#edit` is now able to change any of the SchemaPiece's metadata. (kyranet)
- `Schema#get` (which `SchemaFolder` inherits) now supports paths or arrays. (kyranet)

**Removed**:

- **[BREAKING]** Removed `bwd/*.schema.json` files. (Unseenfaith)
[...]

PR #400, opened 2018-08-15
polish: Schedule and ScheduledTask

kyranet:
### Description of the PR

This PR only polishes the schedule classes and removes old code from Schedule#init (which I forgot to remove in #383, I have checked for any other occurrence of this pattern but I couldn't find any other).

Additionally, this also adds a check for unvalid dates, to prevent Schedule from successfully adding tasks that will be due in a million years (which are displayed as `Invalid Date`).

Also refactors `Schedule#@@iterator` by yielding the array's iterator rather than yielding each value from the array.

Finally, this also makes ScheduleTask.createID much safer by using an incremental number, and fixes ScheduledTask#time's jsdoc type by removing the `?` (it's always a Date instance).

- Removes old code, pulishes and makes some checks more robust

### Semver Classification

### DIFF
--- a/src/lib/schedule/ScheduledTask.js
+++ b/src/lib/schedule/ScheduledTask.js
@@ -233,8 +233,7 @@ class ScheduledTask {
 	 * @private
 	 */
 	static _generateID(client) {
-		i = i < 46656 ? i + 1 : 0;
-		return Date.now().toString(36) + (client.shard ? client.shard.id.toString(36) : '') + i.toString(36);
+		return Date.now().toString(36) + (client.shard ? client.shard.id.toString(36) : '');
 	}
 
 	/**
@@ -252,5 +251,3 @@ class ScheduledTask {
 }
 
 module.exports = ScheduledTask;
-
-let i = 0;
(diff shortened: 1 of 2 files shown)


src/lib/schedule/ScheduledTask.js
```diff
@@ -233,7 +233,8 @@ class ScheduledTask {
 	 * @private
 	 */
 	static _generateID(client) {
-		return Date.now().toString(36) + (client.shard ? client.shard.id.toString(36) : '') + String.fromCharCode((1 % 26) + 97);
+		i = i < 46656 ? i + 1 : 0;
```
bdistin: I don't understand why this is needed. You cannot have 2+ at the same time, except for in the case of shards, at which point we have already handled...

kyranet: With internal sharding, the chances to generate two ids at the same time become greatly bigger, but it's also for the case when two shards generate the same id, which can happen in large scale bots.

bdistin: It's impossible with internal sharding because node is single threaded. Unless internal sharding was done with workers, which it isn't currently.

It's impossible for two shards to generate the same id, because the shard id is taken into account.

PR #402, opened 2018-08-15
SG Auto-Resolving

UnseenFaith:
### Description of the PR
Adds the functionality for creating and resolving settings automatically. This option is disabled by default, and can be enabled via SchemaPiece#resolve. Klasa will automatically resolve any and all resolvable values from the Schema (where possible).

Upon setting resolve to true on any piece in your schema, SG will automatically resolve that piece (and any other pieces that are resolve: true) for you.  During startup, when SG is synced, SG will resolve those values for you and assign them to `<SettingsInstance>[key]`. If for whatever reason, a key is not resolvable, either due to something missing or being deleted from Discord API, SG will set this key to null to keep it consistent with non-resolved values.

- SchemaPiece#resolve
- Base#resolve
- Base#shouldResolve
- SchemaTypes always return full objects and SG will flatten them upon saving to database (subject to change)

### Semver Classification

### DIFF
--- a/src/lib/settings/Settings.js
+++ b/src/lib/settings/Settings.js
@@ -384,7 +384,7 @@ class Settings {
 		if (piece.array) {
 			this._parseArray(piece, route, parsedID, options, result);
 		} else if (this._setValueByPath(piece, parsedID, options.force).updated) {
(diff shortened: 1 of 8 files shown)


UnseenFaith:
Another system has been determined to fill this hole in Settings.

PR #412, opened 2018-08-30
Cleanup

bdistin:
### Description of the PR
Just a pr for others to double check my cleanup work.

-

### Semver Classification


kyranet:
@ShayBox fixed in https://github.com/dirigeants/klasa/pull/412/commits/5efe6914aa300584fcbe9761262a083851be9d96, can you try again, please?

ShayBox:
It works now, Thanks 👍

ShayBox:
I'm guessing this is the place to report any typing problems right now, so:
KlasaClientOptions doesn't include the `production` option

kyranet:
Fixed in https://github.com/dirigeants/klasa/pull/412/commits/b10a52b99a06a2ad89836dbbffdf6f9ba16d38f5, thanks you for reporting this one! I definitely missed this when it landed 😄

ShayBox:
https://github.com/dirigeants/klasa/blob/cleanup/src/monitors/commandHandler.js#L19
This line uses commandText but this line
https://github.com/dirigeants/klasa/blob/cleanup/src/monitors/commandHandler.js#L31
Doesn't return commandText, it does return command though, no commands are registering for me and this is what I tracked it down to.

ShayBox:
The `Cannot read property 'catch' of undefined` issue still persists.
I found that in the typings settings exists on guild and client but not user

ShayBox:
Typings issue, user.settings doesn't exist according to the typings but It does if I log it

bdistin:
Forgive me, I am the first to admit typings aren't something I use/do regularly. Normally kyra's deal. Wouldn't that be handled with https://github.com/dirigeants/klasa/blob/cleanup/typings/index.d.ts#L255 ? Or is something returning d.js's User type that needs to be KlasaUser??

ShayBox:
Ah, you're right, I forgot it's a separate user class, this.client is KlasaClient, message.guild is KlasaGuild, but message.author is a User not KlasaUser, not sure if it's intended to be a KlasaUser or intended to convert to that in code.

bdistin:
Not sure if what I did is the right answer, but can you see if that fixes the typing issue?

ShayBox:
That did, thanks 👍

ShayBox:
Would it be possible to make <GuildMember>.user the KlasaUser class instead of User, or add a KlasaMember/KlasaGuildMember class that's .user is a KlasaUser that I can assign the member type to in params?

kyranet:
We definitely need to make some sort of `discord.js.d.ts` (and make `index.d.ts` import it) file for all extended structures of `discord.js`. So people who extend discord.js/klasa would be able to download it and use it easily. Same for bot developers with a `klasa.d.ts` for consistency.

ShayBox:
I wish It was easier to just add a prop to an existing module's typings without having to re-declare each class and add the existing typings in, it would prevent having to make entirely new classes, but this is mostly a limitation of typescript.

Though it would be nice to have a Klasa version of all the classes from d.js that the usage params can return so all the types are correct straight out of params rather than having to fix the types with `as`
Other then that I don't see any case where the types would be de-synced from the actual code, apart from you fetching something, but you can't really fix that unless you were to override the fetch functions in the typings to return the Llasa variants.

Definitely do that discord.js.d.ts thing, would be interested as I like to redeclare all the discord.js classes with the klasa variants to make things easier myself.

bdistin:
While I can't comment on the feasibility of a `discord.js.d.ts` out of TS ignorance, I do want klasa to not have the type problems you have been having. If that can work, or even help, I am good with the idea. But it will probably need to come in a follow-up pr.

PR #422, opened 2018-09-16
breaking: Start work on new and improved usage strings

KingDGrizzle:
### Description of the PR

This PR is ~~phase one out of two phases of my proposal~~ that will change the way we write our usage strings forever, making the experience easier (in my opinion) and with less bugs too.

This PR completely nukes the concept of `usageDelim`, and instead integrates it into the `usage` string itself.

Take the following usage: `<member:member> [reason:string]`; Before this PR, you'd need to specify usageDelim as " ", and add `[...]` at the end of reason to get it all. Not anymore.

The usage delimiter is now anything between a closed tag and a newly opened one. Yes, that means you can do `<member:member> in <time:duration>`, and you would have to type `@member in 2 days`.

*More in-depth stuff will be found in the guides I'm lazy to write :P*

- **[BREAKING]** Remove `Command#usageDelim`, `Usage#usageDelim` and `Usage#delimitedUsage`

- Others things to be added

### Semver Classification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,6 +125,7 @@ NOTE: For the contributors, you add new entries to this document following this 
 
 ### Changed
 
+- [[422](https://github.com/dirigeants/klasa/pull/422)] **[BREAKING]** Completely change the way usage is parsed, remove `usageDelim`, and change how arguments are returned in the run function.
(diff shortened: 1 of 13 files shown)


bdistin:
This has been executively denied for countless reasons posted in several places, including you can do many of the end goal ideas already, without additional opinionated code (that inevitably breaks other functionality, and ironically the ability to do some of these end goal ideas).

PR #440, opened 2018-09-30
Update documentation

IceeMC:
### Description of the PR
Added the missing options for klasa client.
I also decided to add an example to extend the client for those interested

- Added the unincluded options of KlasaOptions type to GettingStarted.MD
- Added an example of extending the client.

### Semver Classification

### DIFF
--- a/guides/Getting Started/GettingStarted.md
+++ b/guides/Getting Started/GettingStarted.md
@@ -38,25 +38,25 @@ const { Client } = require('klasa');
 
 class SomeExtendedClient extends Client {
 
-    constructor(options) {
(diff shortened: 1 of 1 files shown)


guides/Getting Started/GettingStarted.md
```diff
@@ -30,6 +30,37 @@ new Client({
 }).login('your-bot-token');
 ```
 
+### Extending the client
```
kyranet: I think this should be removed, it's not something related with Klasa, but related with JavaScript. The rest (client options update) looks fine to me.

PR #444, opened 2018-10-01
Removing Unused command Argument In commandInhibited

Skillz4Killz:
### Description of the PR
This PR removes an unused parameter in the commandInhibited event.

Question: If this PR is approved, should I target this PR to `settings` branch to merge with other breaking changes?

- Refactoring commandInhibited

### Note
This **can** be a breaking change for anyone that has transferred the commandInhibited event or commandHandler monitor.

### Semver Classification


bdistin:
Sorry, the intended behavior of this event is to emit the command that is inhibited. Although you can find command on the KlasaMessage object, it's still a part of the interface we want to provide (message, command, reasons)

PR #446, opened 2018-10-02
ReactionHandler check if message.deleted

Tylertron1998:
### Description of the PR

Adds some safety to the ReactionHandler with messages that are deleted/inaccesable.

- Check for message deletes in the ReactionHandler

### Semver Classification


src/lib/util/ReactionHandler.js
```diff
@@ -329,6 +329,7 @@ class ReactionHandler extends ReactionCollector {
 	 * @private
 	 */
 	async _queueEmojiReactions(emojis) {
+		if (this.message.deleted) this.stop();
```
bdistin: needs to `return this.stop();`

PR #455, opened 2018-10-16
fullUsage consistency & improvement

tech6hutch:
### Description of the PR
Fixes inconsistencies in and improves CommandUsage#fullUsage.

- changed to slice the prefix regardless of whether it's an array
  - fixes inconsistency where, with the prefixCaseInsensitive option on, the exact case is only preserved if prefix is array
  - fixes inconsistency where, when the user used the mention prefix, it would show the first prefix if prefix is array, or `<@bot ID here>` if prefix is not array
- changed to show the mention prefix the way users actually use it, e.g. `@username#0000`

### Semver Classification


bdistin:
This requires some simplification.

tech6hutch:
@bdistin is that good?

bdistin:
flip the ifs, and make the newly second if an else if.

PR #456, opened 2018-10-17
Changes to commandHandler's events

tech6hutch:
### Description of the PR
There are two parts to this PR: new parameters for the commandUnknown event, and a new event for messages without _any_ command, valid or invalid. Klasa currently lacks any substitute for these things, beyond transferring and editing the commandHandler monitor.

#### New Arguments to commandUnknown

The commandUnknown event is lacking some vital information. Existing arguments:
  - `message`: Since there was no valid command, this Message does not have a prefix or command registered to it.
  - `command`: The name of the nonexistent command. This can't be reliably used to get the text after the command name with String#indexOf, since it's possible that the prefix could contain the command name.

As it is now, there is no reliable way to get the "arguments" given for the nonexistent command (the text after the command name), without re-parsing the prefix. Therefore, I propose new arguments to the event:
  - `prefix`: The prefix that commandHandler found in the message.
  - `prefixLength`: Useful for slicing off the prefix. You can get the "arguments" to the command with:
```js
const args = message.content
  .substring(prefixLength).trim()
  .substring(command.length).trim();
```

These new arguments make unknown command handling much more robust.

#### New commandlessMessage Event

However, there's another gap: from a monitor, there's no way to tell whether the message does or does not have a command. Testing for `message.command` would be a race condition. Even if it wasn't, there's no way to tell whether the commandUnknown event was triggered, or whether the message begins with a prefix, without redoing those checks yourself.

For this reason, I propose a new "commandlessMessage" event, for messages that do not have a parse-able command. Its arguments:
  - `message`
  - `prefix`: If any. Can be easily used to determine whether the message began with a prefix: `if (prefix)`, `if (!prefix)`, etc.
  - `prefixLength`: If any.

This event can be useful for monitoring messages that you can be sure the commandHandler hasn't already..._handled_.

### Alternate Ideas

Options for monitors to `ignorePrefixed` and `ignoreCommands` would mostly solve the commandlessMessage problem, but this would require some restructuring. For `ignorePrefixed`, prefix parsing would have to be moved out of commandHandler and into the message event. For `ignoreCommands`, some parsing of commands would have to happen outside of commandHandler, too, which is unpleasant.

On the plus side, this would allow the prefix and prefixLength (if there was any prefix) to be passed to monitors.

- added new commandlessMessage event, for messages that don't have commands, not even unknown ones
- added prefix and prefixLength args to commandUnknown event

### Semver Classification

- [X] This PR adds methods or properties to the framework interface.


kyranet:
Additionally, this could refactor commandHandler a little bit plus allow one step further of customization if you make a core commandUnknown event and move the prefix reminder there. Right now it would need to transfer the commandHandler monitor to remove or customize it.

tech6hutch:
@kyranet the prefix reminder would have to use the commandlessMessage event. I'm worried that the difference between the two events might be a little confusing. I suppose I could combine them, just passing `null` or whatever if there's no commandText.

kyranet:
Oh, you're right, my bad, I didn't read well you were adding a new event. I guess we should wait for @bdistin

tech6hutch:
For an example of how these might be used, see my bot.

commandHandler is mostly the same as in this PR: https://github.com/tech6hutch/MissyBot/blob/master/src/monitors/commandHandler.js (however I also modified it to ignore punctuation in command names)
commandlessMessage: https://github.com/tech6hutch/MissyBot/blob/master/src/events/commandlessMessage.js
commandUnknown: https://github.com/tech6hutch/MissyBot/blob/master/src/events/commandUnknown.js

UnseenFaith:
This needs some more internal thinking.. we're really starting to blur the line between the monitors and commands.

bdistin:
commandlessMessage is a no from me. the new commandUnknown needs docs and typings

kyranet:
Bump, please apply @bdistin's requested changes. @tech6hutch

tech6hutch:
@bdistin I did your requested changes, including types and docs. Let me know if I missed anything.

PR #467, opened 2018-10-27
refactor: Convert GatewayDriver into a Collection<string, Gateway>

kyranet:
### Description of the PR

Dictionary-type classes are a bad practise in JavaScript, even though Settings is a dictionary class too (with the difference it doesn't change its properties on runtime but are created on instantiation). This PR refactors GatewayDriver to follow JavaScript's best practises.

This is a large breaking change for anybody using SG, with the only change:

```javascript
this.client.gateways.guilds;
```

Becomes:

```javascript
this.client.gateways.get('guilds');
```

- **[BREAKING]** Changed `GatewayDriver` to extend `Collection<string, Gateway>` as opposed to being a dictionary type.

### Semver Classification

### DIFF
--- a/src/lib/settings/GatewayDriver.js
+++ b/src/lib/settings/GatewayDriver.js
@@ -61,7 +61,7 @@ class GatewayDriver extends Collection {
 	register(name, { provider = this.client.options.providers.default, schema = new Schema() } = {}) {
 		if (typeof name !== 'string') throw new TypeError('You must pass a name for your new gateway and it must be a string.');
 		if (!(schema instanceof Schema)) throw new TypeError('Schema must be a valid Schema instance.');
-		if (this.has(name)) throw new Error(`The key '${name}' is either taken by another Gateway or reserved for GatewayDriver's functionality.`);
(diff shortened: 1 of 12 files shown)


src/lib/settings/GatewayDriver.js
```diff
@@ -92,13 +61,11 @@ class GatewayDriver {
 	register(name, { provider = this.client.options.providers.default, schema = new Schema() } = {}) {
 		if (typeof name !== 'string') throw new TypeError('You must pass a name for your new gateway and it must be a string.');
 		if (!(schema instanceof Schema)) throw new TypeError('Schema must be a valid Schema instance.');
-		if (this.name !== undefined && this.name !== null) throw new Error(`The key '${name}' is either taken by another Gateway or reserved for GatewayDriver's functionality.`);
+		if (this.has(name)) throw new Error(`The key '${name}' is either taken by another Gateway or reserved for GatewayDriver's functionality.`);
```
UnseenFaith: Error is wrong, can only be taken by another gateway now that it'ss powered by map
```suggestion
		if (this.has(name)) throw new Error(`The key '${name}' is taken by another Gateway.`);
```

PR #470, opened 2018-10-30
fix: Replace process.binding usage with native v8

KingDGrizzle:
### Description of the PR

Node.JS is deprecating `process.binding`, which was used in Type to fetch info about a Promise. 
This PR replaces it with native v8 calls because v8.

-

### Semver Classification

### DIFF
--- a/src/lib/util/Type.js
+++ b/src/lib/util/Type.js
@@ -4,7 +4,7 @@ const util = require('./util');
 
 setFlagsFromString('--allow-natives-syntax');
 
-const getPromiseDetails = runInThisContext('p => p instanceof Promise ? { status: %PromiseStatus(p), result: %PromiseResult(p) } : []');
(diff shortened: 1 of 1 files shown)


src/lib/util/Type.js
```diff
@@ -1,5 +1,5 @@
-const util = require('./util');
-const { getPromiseDetails } = process.binding('util');
+require('v8').setFlagsFromString('--allow-natives-syntax');
+const getPromiseDetails = require('vm').runInThisContext('p => p instanceof Promise ? { status: %PromiseStatus(p), result: %PromiseResult(p) } : null');
```
kyranet: The code misses a semicolon:
```suggestion
const getPromiseDetails = require('vm').runInThisContext(
	'p => p instanceof Promise ? { status: %PromiseStatus(p), result: %PromiseResult(p) } : null;');
```

PR #471, opened 2018-11-01
Settings rewrite

kyranet:
### Description of the PR

This PR changes the way Settings work from almost the very ground up. It's an all-ways breaking change for everything that relies on SettingsGateway (in other words, this breaks almost every single bot that relies on SettingsGateway).

For starters, Settings will not longer be objects, but (nested) extended maps.

**Added**:

- Added `Gateway#create` and `Gateway#acquire`.
- Added `SETTING_GATEWAY_CHOOSE_KEY` and `SETTING_GATEWAY_UNCONFIGURABLE_FOLDER` i18n keys into en-US.
- Added `SettingsFolder`, with the common operations (update, reset...) for better OOP compliancy.

**Changed**:

- Modified `Schema#defaults` type from Object literal to a `SettingsFolder` instance.
- Modified `Schema#defaults` to be a property instead of a getter.
- Modified `Gateway#get` to take only id. For get or create, use `Gateway#acquire` instead.
- Modified `Settings#{update,reset}` output to return `{ key: string, value: any, piece: Schema | SchemaPiece }` instead of `{ data: [string, any], piece: Schema | SchemaPiece }` in the `updated` field.
- Modified `Settings` completely, they're not longer dictionaries but (nested) `SettingsFolder`s.
- Modified `Util.resolveGuild` to also resolve members.
- Modified initialization order to prepare `Schedule` before initializing pieces.
- Modified Schema#get overload to not check safely to reduce overhead.
- Modified Settings#sync to force sync if the entry was not synchronized.
- Renamed `Gateway#syncQueue` to `Gateway#syncMap`, changing also its type from `Collection<string, Promise<Settings>>` to `WeakMap<string,-  Promise<Settings>>`.
- Renamed `GatewayDriver#type` to `GatewayDriver#name`.
- Renamed `SettingsUpdateOptions.{action,arrayPosition,avoidUnconfigurable}` to `SettingsUpdateOptions.{arrayAction,arrayIndex,onlyConfigurable}`.

**Removed**:

- Removed `Gateway#defaults`. Refer to `Schema#defaults` instead.

### Semver Classification


src/monitors/commandHandler.js
```diff
@@ -14,7 +14,7 @@ module.exports = class extends Monitor {
 	async run(message) {
 		if (message.guild && !message.guild.me) await message.guild.members.fetch(this.client.user);
 		if (!message.channel.postable) return undefined;
-		if (this.mentionOnly.test(message.content)) return message.sendLocale('PREFIX_REMINDER', [message.guildSettings.prefix.length ? message.guildSettings.prefix : undefined]);
```
bdistin: the length check is for empty prefix arrays

kyranet:
Is there anything else missing or can we merge this?

cfanoulis:
So, how would this work? Like `settings.get("id`?

kyranet:
> So, how would this work? Like `settings.get("id")`?

`settings.id` is still available as it's not part of the schema itself, but anything that is in the schema is stored as k-v in the Map. You have to do `settings.get('prefix')` to get the prefix's value, but inside folders, to prevent `settings.get('roles').get('administrator')`, we have overloaded the get method to get recursively: `settings.get('roles.administrator')`.

It's more like `settings.path.to.key` -> `settings.get('path.to.key')`

cfanoulis:
Oh, nice thank you.

PR #472, opened 2018-11-03
docs: Fixed ReactionHandler time descriptions

tech6hutch:
### Description of the PR
~Collector, which ReactionCollector extends, which ReactionHandler extends, has a `time` option, but it doesn't set it as a property. ReactionHandler sets this as a property, and attempts to set a default time, but it won't do any good after `super()` has already run, since Collector has already set the timeout based on `options.time` and not a property. In addition, this (attempted) default was not documented. This PR corrects these things.~

I fix doc.

- Fix doc

### Semver Classification

### DIFF
--- a/src/lib/util/ReactionHandler.js
+++ b/src/lib/util/ReactionHandler.js
@@ -33,7 +33,6 @@ class ReactionHandler extends ReactionCollector {
 	 * @param {Emoji[]} emojis The emojis which should be used in this handler
 	 */
 	constructor(message, filter, options, display, emojis) {
-		if (!options.time) options.time = 30000;
 		super(message, filter, options);
 
 		/**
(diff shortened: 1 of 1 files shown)


bdistin:
Although the docs are wrong, that's not what this.time is for... see pr #57

tech6hutch:
Oh, my apologies. I will fix that.

tech6hutch:
@bdistin I corrected my (in)corrections. This is probably out of scope for this PR, but would you support renaming the property from `time` to something like `menuTime`? The current name is misleading, especially for people who haven't used the Discord.js ReactionCollector before and might assume, from the name, that this option is used for general collector timeout.

bdistin:
docs still wrong, this.time is used in the jump menu prompt

PR #473, opened 2018-11-03
deps: Update ping command for Internal Sharding

kyranet:
### Description of the PR

What the title says, Client#{ping,pings} was moved to WebSocketManager#ping

- Updated klasa for IS

### Semver Classification


bdistin:
More needs to be done to be compatible with d.js's internal sharding.

client.shard.id can be an array now:

shardID portions of the log, should be moved to commandLogging (since all bots will have message.guild ? message.guild.shardID : 0); meaning client can be removed from KlasaConsole. Else the shard id may show a list of shard ids.

All broadcast evals need to be re-evaluated as comparing `this.shard.id === ${this.client.shard.id}` will be eranious when the client.shard.id is an array.

PR #497, opened 2018-11-20
Removing Extendable Old SelfBot Code

Skillz4Killz:
### Description of the PR
Removes GroupDMChannel in extendables which is used by selfbots

-

### Semver Classification


tech6hutch:
It hasn't been removed from Discord.js because it could be used in the future...or something

bdistin:
That's not necessarily userbot code, as groupdms could be allowed for bots at any time. The same reason why d.js hasn't removed groupdms from it's lib.

PR #526, opened 2018-12-03
more descriptive msg for nsfw cmds inhibited

gc:
### Description of the PR

"You may not use NSFW commands in this channel." doesn't tell the user why they cannot use the command and/or how to fix the issue so they can use it in this channel

"This command might post NSFW content, please mark this channel as NSFW to use it here." - tells the user that it might output NSFW content, and that they can mark the channel as NSFW to use it there.

-

### Semver Classification

- [X] This PR fixes a bug and does not change the (intended) framework interface.

### DIFF
--- a/src/languages/en-US.js
+++ b/src/languages/en-US.js
@@ -57,7 +57,7 @@ module.exports = class extends Language {
 			INHIBITOR_COOLDOWN: (remaining) => `You have just used this command. You can use this command again in ${remaining} second${remaining === 1 ? '' : 's'}.`,
 			INHIBITOR_DISABLED: 'This command is currently disabled.',
 			INHIBITOR_MISSING_BOT_PERMS: (missing) => `Insufficient permissions, missing: **${missing}**`,
-			INHIBITOR_NSFW: 'You may not use NSFW commands in this channel.',
+			INHIBITOR_NSFW: 'This command might post NSFW content, please mark this channel as NSFW to use it here.',
 			INHIBITOR_PERMISSIONS: 'You do not have permission to use this command.',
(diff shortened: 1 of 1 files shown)


Tylertron1998:
"You may not use NSFW commands in a non-NSFW channel." is still as concise, but clearer imo.

bdistin:
"You can only use NSFW commands in NSFW channels."

kyranet:
I think "You cannot use NSFW commands in SFW channels." would be better, without the word duplication.

bdistin:
I don't see the duplication of NSFW as a problem since that's how it's defined as a setting both in klasa and in discord. However, I do think it sounds funny `You cannot use not ...`

KingDGrizzle:
`You may not use NSFW Commands in channels that aren't marked as NSFW` or similar 😄

bdistin:
The problem with tyler/vlad's suggestion is they are double negatives, which should be avoided here.

Skillz4Killz:
Solves the duplicate NSFW plus the double negative + solves the OP concern about telling the user to edit the settings.

`This command will not work until this channel allows NSFW content in its settings.`

gc:
I still think my original one is best as it: points out that the command is NSFW and that it can post nsfw content, and that you have to mark the channel as NSFW to use it. Every other suggestion in the thread doesn't do the first thing (explicitly state that the command posts NSFW things) - which I think its better if it does: `"This command might post NSFW content, please mark this channel as NSFW to use it here."`. I'd say it could be objectively better because it provides more information as opposed to being subjectively better because it sounds better or is shorter/succinct.

Although I suggest Ao just picks the one he likes best as we could discuss this forever :stuck_out_tongue_closed_eyes:

bdistin:
It's irrelevant to speculate on what the command may or may not post. That's why none of the alternatives state that... The message is meant to indicate the command's not going to work "here".

PR #539, opened 2018-12-11
Allows Inhibitors to have sync run methods

bdistin:
### Description of the PR
I also thought this was already done this way....

-

### Semver Classification

### DIFF
--- a/src/lib/structures/Inhibitor.js
+++ b/src/lib/structures/Inhibitor.js
@@ -33,6 +33,22 @@ class Inhibitor extends Piece {
 	}
 
 	/**
+	 * The async wrapper for running inhibitors
+	 * @since 0.5.0
+	 * @param {KlasaMessage} message The message that triggered this inhibitor
+	 * @param {Command} command The command to run
+	 * @returns {(void|string)}
+	 * @private
+	 */
+	async _run(message, command) {
+		try {
+			return await this.run(message, command);
+		} catch (err) {
+			return err;
+		}
+	}
+
+	/**
 	 * The run method to be overwritten in actual inhibitors
 	 * @since 0.0.1
 	 * @param {KlasaMessage} message The message that triggered this inhibitor
--- a/src/lib/structures/InhibitorStore.js
+++ b/src/lib/structures/InhibitorStore.js
@@ -26,7 +26,19 @@ class InhibitorStore extends Store {
 	 */
 	async run(message, command, selective = false) {
 		const mps = [];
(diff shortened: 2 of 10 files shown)


src/lib/structures/InhibitorStore.js
```diff
@@ -26,7 +26,19 @@ class InhibitorStore extends Store {
 	 */
 	async run(message, command, selective = false) {
 		const mps = [];
-		for (const inhibitor of this.values()) if (inhibitor.enabled && (!selective || !inhibitor.spamProtection)) mps.push(inhibitor.run(message, command).catch(err => err));
+
+		for (const inhibitor of this.values()) {
```
kyranet: For consistency with other structures, each call should be done in another method, to avoid re-creating a local function on every call, making inhibitors more expensive than what they currently are.

For example, MonitorStore:

https://github.com/dirigeants/klasa/blob/3bb3000dafac8a35470b6d6a6c8b4847cca66f6a/src/lib/structures/MonitorStore.js#L19-L41

bdistin: shoot, you are right...

kyranet: Additionally, I'd like to discriminate throws with string, and throws with the Error constructor, so when there's an inhibitor throwing a legit error (e.g. breaking changes affecting one inhibitor), it doesn't get sent to the channel without logging any of the stack trace. Though this would make this PR semver-major...

src/lib/structures/Inhibitor.js
```diff
@@ -32,6 +32,22 @@ class Inhibitor extends Piece {
...
 
+	/**
+	 * The async wrapper for running inhibitors
+	 * @since 0.5.0
+	 * @param {KlasaMessage} message The message that triggered this inhibitor
+	 * @param {Command} command The command to run
+	 * @returns {(void|string)}
+	 * @private
+	 */
+	async _run(message, command) {
```
kyranet: Oh yeah, this is way better as it allows better customization (like making our own Inhibitor class extending this one to always log `Error` instances in an `inhibitorError` event), just typings and this should be fine to merge 👍

kyranet: Also, should we make this `protected` in typings? So we can override its behaviour. Not sure if you want to change `@private` to `@protected`.

bdistin: not sure if protected is handled by our jsdocs/website (atleast properly). It can be protected in typings.

kyranet: No, they aren't handled at the moment. `Store#registerCoreDirectory` is marked with `@protected`. https://klasa.js.org/#/docs/klasa/master/class/Store?scrollTo=registerCoreDirectory but the documentation does not show any label for it.

PR #545, opened 2018-12-13
QueryBuilder rewrite

kyranet:
### Description of the PR

~~The `settings` branch already broke the SQL providers to apply a bugfix, so let's break them even more now that we are on it.~~

This PR enhances the QueryBuilder's interface a lot by giving it a much friendlier interface extending Map and using an `add` method that processes each datatype correctly, as opposed to having 2 interfaces mixed and having to de-mix them into two objects based on keys given from the constants, which is very messy and slow.

Additionally, QueryBuilder's internals were overcomplicated, the replacement to use Maps makes this really easier to use and understand.

This PR also enhances docs and typings for better understanding while reducing the lines of code spent for documentation (functions were duplicated but without the use of arguments, resulting on documentation data loss).

To show the before and now difference... before:

```javascript
this.qb = new QueryBuilder({
	boolean: 'BOOL',
	integer: ({ max }) => max >= 2 ** 32 ? 'BIGINT' : 'INTEGER',
	float: 'DOUBLE PRECISION',
	uuid: 'UUID',
	json: { type: 'JSON', resolver: (input) => `'${JSON.stringify(input)}'::json` },
	any: { type: 'JSON', resolver: (input) => `'${JSON.stringify(input)}'::json` },
	array: type => `${type}[]`,
	arrayResolver: (values, piece, resolver) =>
		values.length ? `array[${values.map(value => resolver(value, piece)).join(', ')}]` : "'{}'",
	formatDatatype: (name, datatype, def = null) =>
		`"${name}" ${datatype}${def !== null ? ` NOT NULL DEFAULT ${def}` : ''}`
});
```

Now:

```javascript
this.qb = new QueryBuilder({
	array: type => `${type}[]`,
	arrayResolver: (values, piece, resolver) =>
		values.length ? `array[${values.map(value => resolver(value, piece)).join(', ')}]` : "'{}'",
	formatDatatype: (name, datatype, def = null) =>
		`"${name}" ${datatype}${def !== null ? ` NOT NULL DEFAULT ${def}` : ''}`
})
	.add('boolean', { type: 'BOOL' })
	.add('integer', { type: ({ max }) => max >= 2 ** 32 ? 'BIGINT' : 'INTEGER' })
	.add('float', { type: 'DOUBLE PRECISION' })
	.add('uuid', { type: 'UUID' })
	.add('any', { type: 'JSON', resolver: (input) => `'${JSON.stringify(input)}'::json` })
	.add('json', { extends: 'any' });
```

- Changed the entire API for `QueryBuilder` to extend a Map instead of being a pseudo-map.

### Semver Classification


kyranet:
@bdistin @UnseenFaith Should we remove this function and move it to a serializer option in QueryBuilder? https://github.com/dirigeants/klasa/blob/4bafae48c5f7946ac9fa98f778b5380616d80f39/src/lib/structures/SQLProvider.js#L111-L150

PR #373, opened 2018-07-25
locale descriptions and extend helps get languages

bdistin:
### Description of the PR
instead of messages since messages are not always available.

### Changes Proposed in this Pull Request (List new items in CHANGELOG.MD)

-

### Semver Classification

- [ ] This PR only includes documentation or non-code changes.
- [ ] This PR fixes a bug and does not change the (intended) framework interface.
- [ ] This PR adds methods or properties to the framework interface.
- [x] This PR removes or renames methods or properties in the framework interface.

PR #565, opened 2019-01-18
Making Help Great Again!

Skillz4Killz:
### Description of the PR
This PR allows any developer to be able to use anything on the message object inside the help functions. Currently, the only thing being passed in is the `message.language` however that is incredibly limiting. This makes it almost impossible to do some really useful things in help command.

As an example if I wanted to show the users how to use a command by showing examples:

`**.command x y z** Will do xyz`
`**.command a b c*** Will do abc`

This doesn't actually work because you are unable to get the prefix to show the user. But wait, there is more!

- Try using guild settings. IMPOSSIBLE
- Try using user settings. IMPOSSIBLE.

**.command x y z** This will change your current XYZ to xyz

The list of things that are not possible with this current implementation just keep going. It is extremely limiting and not flexible whatsoever. 

This PR allows passing the entire message object which can be deconstructed to provide the variables you wish or just message itself if you wish.

```js
description: ({ language }) => language.......
extendedHelp: ({ language, guild, guildSettings, client, user }) => language.......

description: message => message.language......
extendedHelp: message => message.language.....
```

The amount of flexibility that this provides is so much more than only providing a language.

- Passes in `message` instead of `language` to help description and extendedHelp

### Semver Classification

### DIFF
--- a/src/commands/Admin/blacklist.js
+++ b/src/commands/Admin/blacklist.js
@@ -6,7 +6,7 @@ module.exports = class extends Command {
 	constructor(...args) {
 		super(...args, {
 			permissionLevel: 10,
-			description: language => language.get('COMMAND_BLACKLIST_DESCRIPTION'),
+			description: ({ language }) => language.get('COMMAND_BLACKLIST_DESCRIPTION'),
 			usage: '<User:user|Guild:guild|guild:str> [...]',
 			usageDelim: ' ',
 			guarded: true
--- a/src/commands/Admin/conf.js
+++ b/src/commands/Admin/conf.js
@@ -8,7 +8,7 @@ module.exports = class extends Command {
 			permissionLevel: 6,
 			guarded: true,
 			subcommands: true,
-			description: language => language.get('COMMAND_CONF_SERVER_DESCRIPTION'),
+			description: ({ language }) => language.get('COMMAND_CONF_SERVER_DESCRIPTION'),
 			usage: '<set|show|remove|reset> (key:key) (value:value) [...]',
 			usageDelim: ' '
 		});
--- a/src/commands/Admin/disable.js
+++ b/src/commands/Admin/disable.js
@@ -6,7 +6,7 @@ module.exports = class extends Command {
 		super(...args, {
 			permissionLevel: 10,
 			guarded: true,
-			description: language => language.get('COMMAND_DISABLE_DESCRIPTION'),
+			description: ({ language }) => language.get('COMMAND_DISABLE_DESCRIPTION'),
 			usage: '<Piece:piece>'
 		});
 	}
--- a/src/commands/Admin/enable.js
+++ b/src/commands/Admin/enable.js
@@ -6,7 +6,7 @@ module.exports = class extends Command {
 		super(...args, {
 			permissionLevel: 10,
 			guarded: true,
-			description: language => language.get('COMMAND_ENABLE_DESCRIPTION'),
+			description: ({ language }) => language.get('COMMAND_ENABLE_DESCRIPTION'),
 			usage: '<Piece:piece>'
 		});
 	}
--- a/src/commands/Admin/eval.js
+++ b/src/commands/Admin/eval.js
@@ -8,8 +8,8 @@ module.exports = class extends Command {
 			aliases: ['ev'],
 			permissionLevel: 10,
 			guarded: true,
-			description: language => language.get('COMMAND_EVAL_DESCRIPTION'),
-			extendedHelp: language => language.get('COMMAND_EVAL_EXTENDEDHELP'),
+			description: ({ language }) => language.get('COMMAND_EVAL_DESCRIPTION'),
+			extendedHelp: ({ language }) => language.get('COMMAND_EVAL_EXTENDEDHELP'),
 			usage: '<expression:str>'
 		});
 	}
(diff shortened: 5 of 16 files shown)


pollen5:
This was like this before but got removed in a breaking change, see PR #373 for more info.

Skillz4Killz:
That is very interesting. Not really sure why a `message` would be unavailable when `language` is passed in by doing `message.language`. The language is literally passed in from a message. Very curious 🤔

vladfrangu:
> That is very interesting. Not really sure why a message would be unavailable when language is passed in by doing message.language. The language is literally passed in from a message. Very curious 🤔 Also how did the command run in the first place if message is not available.

One function. Command#toJSON. And that says it all. Emulating a fake message object with all the info is basically impossible

Skillz4Killz:
@vladfrangu When is Command.toJSON() run that it would make this make the message invalid. I know for a fact it does not cause an issue with this because this is working. Could you show me how that function causes a problem in generating a help command please

kyranet:
#373 is a bugfix PR that fixed `Command#toJSON()` not working correctly since it was passing a `Language` instance, and not a `KlasaMessage` instance. This PR re-implements the same bug and breaks `klasa-dashboard-hooks`. Plus what @vladfrangu said, and what @bdistin already said repeatedly in the official guild during the past.

Feel free to think of another way that does not break existing applications and or bots, I do not think the old behaviour will be accepted anymore.

Skillz4Killz:
@kyranet Ahh so the problem is `klasa-dashboard-hooks` which breaks it. That is very interesting indeed. But why remove this amazing flexibility and functionality for every bot for only those that want to use `klasa-dashboard-hooks`? 

The ideal solution would be to create a plugin that when you use `klasa-dashboard-hooks` overrides this behavior and allows command.toJSON to work. That way anyone that wants to use klasa-dashboard-hooks won't have an issue and in the meantime, anyone that doesn't want to use it can use this better method.

`Instead everything should be an abstract working base that people can personalize themselves to their own needs.` Language only is very limiting and not at all as flexible or personalizable as a message.

`Everything in core should be generally useful for the majority of Klasa bots. ` How many bots actually use `klasa-dashboard-hooks` as compared to bots that don't? I would say the majority do not use it.

`As much of the framework as possible is meant to be customizable for any possible use-case, even if the use-case is niche.` Customizable for any use-case. Language breaks that.

I don't think this current behavior follows the Klasa guidelines at all.

kyranet:
> @kyranet Ahh so the problem is `klasa-dashboard-hooks` which breaks it. That is very interesting indeed. But why remove this amazing flexibility and functionality for every bot for only those that want to use `klasa-dashboard-hooks`?

We did not remove any flexibility, we just fixed a bug by making it work the way it was expected to since the introduction of language files.

> The ideal solution would be to create a plugin that when you use `klasa-dashboard-hooks` overrides this behavior and allows command.toJSON to work. That way anyone that wants to use klasa-dashboard-hooks won't have an issue and in the meantime, anyone that doesn't want to use it can use this better method.

Rule 1# of official plug-ins, they must not modify the existing behaviour of `klasa` or `discord.js`.

> `Instead everything should be an abstract working base that people can personalize themselves to their own needs.` Language only is very limiting and not at all as flexible or personalizable as a message.

Getting a `KlasaMessage` itself is also very limiting, before #373 landed, we were "forging" fake messages (which, by the way, slow down and deoptimize Klasa's internals) as a workaround, just to be able to show the translated messages in our websites. This PR will make us re-introduce the same workaround again, which will make the code worse. And this workaround should not belong to core, since as @vladfrangu described, you may want to access to a property that is not available (like author.tag, what is author here?).

> `Everything in core should be generally useful for the majority of Klasa bots. ` How many bots actually use `klasa-dashboard-hooks` as compared to bots that don't? I would say the majority do not use it.

Probably more than the bots that use message.author or any other property in their help messages.

> I don't think this current behavior follows the Klasa guidelines at all.

It actually does. The behaviour you want to introduce does not.

Skillz4Killz:
I think your getting lost on the main point that I am trying to make is that the current method is not something that should be in core. When you fixed this bug  the solution you used actually removed all this flexibility and made it only for those that use KDH and taking away from every other klasa user. 

The plugin idea is just my solution I am sure you all can think of a better solution together than the one I came up with. 

Once again faking a KlasaMessage is only for users of KDH not for klasa core users. Passing a message and not language is only limiting for KDH users not every single user of klasa. My solution in this PR works for any bot that does not use KDH. At the end of the day the users that use KDH are probably more advanced then any average Klasa user and the advanced users have the ability to transfer the commands and alter it from passing `message` to `message.language`. For the niche user that needs this behavior, they should be transferring it and doing it the way they need it just like all other cases. Klasa core should always be built for the majority use case.

`Probably more than the bots that use message.author or any other property in their help messages.` This is very bad logic kyra. First, my question was entirely different and was not answered. `How many bots actually use klasa-dashboard-hooks as compared to bots that don't?` My question is how many bots use KDH versus those that do not. Majority of bots, in my opinion, would NOT use KDH and therefore this behavior goes against the core guidelines of Klasa. Second,  klasa is restricting them from doing so. You don't let the majority of users even consider the possibility of using them because they dont even have access to it since you only pass language.

In my opinion, changing Klasa core behavior for a niche use case that takes away a lot of use cases for the majority of users is never a good solution. The current solution goes against the klasa guidelines in so many ways.

image

PR #583, opened 2019-02-11
typings: Fixed an error, lowered strictness for Command#run's return

kyranet:
### Description of the PR

What the title says. `Constructable<any>` works, but only for non-abstract classes (so if we wanted to extend klasa's structures, which are abstract, we would get an error).

- Fixed an error, lowered strictness for Command#run's return in typings

### Semver Classification


tech6hutch:
So finalizers aren't guaranteed to get a KlasaMessage anymore? Not that they were completely guaranteed before, I suppose...

kyranet:
No, they weren't, @tech6hutch.

Finalizers receive as `response` whatever the command ran returned.

PR #593, opened 2019-02-23
fix some partials related things with Arguments

bdistin:
### Description of the PR
Fixes the Member argument's behavior if the cached member is a partial. Adds actual fetching capability to the Channel argument (to auto-resolve any partials issues), and supports DM Channels even if they aren't cached or if you use a user mention.

-

### Semver Classification

### DIFF
--- a/src/arguments/channel.js
+++ b/src/arguments/channel.js
@@ -4,10 +4,10 @@ module.exports = class extends Argument {
 
 	async run(arg, possible, message) {
 		// DM Channel support
-		const user = this.constructor.regex.userOrMember.test(arg) ? await this.client.users.fetch(this.constructor.regex.userOrMember.exec(arg)[1]) : null;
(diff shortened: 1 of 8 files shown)


src/arguments/channel.js
```diff
@@ -2,8 +2,12 @@ const { Argument } = require('klasa');
 
 module.exports = class extends Argument {
 
-	run(arg, possible, message) {
-		const channel = this.constructor.regex.channel.test(arg) ? this.client.channels.get(this.constructor.regex.channel.exec(arg)[1]) : null;
+	async run(arg, possible, message) {
+		// DM Channel support
+		const user = this.constructor.regex.userOrMember.test(arg) ? await this.client.users.fetch(this.constructor.regex.userOrMember.exec(arg)[1]) : null;
```
vladfrangu: What if you pass in an ID and this check throws? Say I pass in an ID for a channel, which matches the regex for userOrMember, users.fetch WILL throw

If anything this should be merged down into one ternary, if possible

bdistin: oh, the catch(() => null)s are gone, need to double check all args

PR #615, opened 2019-03-09
README TUTORIAL switch to master from stable 

Anish-Shobith:
### Description of the PR
In the readme when you click the tutorial it redirects you to stable documentation but we should be using master as stable is no longer in use 
Changed from stable to master in the readme file
### Semver Classification

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Following suit from Komada (the Croatian word for "pieces"), Klasa is the Croati
 
 ## Time to get started!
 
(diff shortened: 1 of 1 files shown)


bdistin:
This change is wrong. Even though this library is in development mode because it has been written for d.js v12 since day one, and that stable was long since broken by d.js's instability; it's still correct for these links to point to stable because that's where stability will be when klasa v1.0.0 and d.js v12.0.0 are released as stable. We aren't going to change things for now in all of these random places to have to track down all of these places to revert these types of changes when things will inevitably change in the future.

Anish-Shobith:
Oh i see

PR #688, opened 2019-05-15
rename event files that have core klasa functionality

### LINKED ISSUE #432: Proposal: Non overriding events
### Describe your proposal
An event option to prevent an event from overriding the built-in core event if provided.

### Use-Cases for your proposal
You could have both the core event functional, and a personal event without one being overridden.

An event with the option `override: false` will not override a core event.

### Further details
By default all events's override option would default to true, to keep the current functionality the same and not break things, and allow users to create events that do not override core events, Ex.
The core message event working
A personal message event with `override: false`, doing something completely different than the core event.

This keeps the core events in-framework, keeps them up to date with the framework, rather than having to require users to copy all the code from events they need access to into their events then keep that code up to date, when they never modify the code in the first place.

Another possible configuration would be, Ex.
A message event with `override: true` (default) that overrides the core event, that you would copy and modify the core event's code into, to modify the functionality,
Then allow you to add a second event with `override: false` that has an entirely different use, doesn't contain the core event code or any variant, and both of them function.

- [x] This (as described) doesn't include any breaking changes.

bdistin:
### Description of the PR
#432 Breaking, renames event files with core Klasa functionality to include a core prefix.

The decision if `core` or not came down to if a Klasa feature would break if the file was overridden with a blank event. Defining Klasa feature as command editing, guild blacklist, etc; and not logging/error handling. The idea being if someone is making their own debug event they want to overwrite the debug functionality. But if you make a message event, all monitors must still function.

If you disagree with any of the event files being renamed or not renamed, please comment your reasoning why said event should or shouldn't be prefixed.

-

### Semver Classification

### DIFF
--- a/src/events/coreGuildCreate.js
+++ b/src/events/coreGuildCreate.js
@@ -2,6 +2,10 @@ const { Event } = require('klasa');
 
 module.exports = class extends Event {
 
+	constructor(...args) {
+		super(...args, { event: 'guildCreate' });
+	}
+
 	run(guild) {
 		if (!guild.available) return;
 		if (this.client.settings.guildBlacklist.includes(guild.id)) {
--- a/src/events/coreGuildDelete.js
+++ b/src/events/coreGuildDelete.js
@@ -2,6 +2,10 @@ const { Event } = require('klasa');
 
 module.exports = class extends Event {
 
+	constructor(...args) {
+		super(...args, { event: 'guildDelete' });
+	}
+
 	run(guild) {
 		if (this.client.ready && guild.available && !this.client.options.preserveSettings) guild.settings.destroy().catch(() => null);
 	}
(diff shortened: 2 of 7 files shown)


Skillz4Killz:
I don't think `settingsUpdateEntry.js` needs to be renamed. 

1) This event is not a discord.js event or any event a normal user would ever be making on their own. 
2) The only reason a user would make this event is that they have a custom sharding system. Either way this file is named that user would have to transfer this file to disable this form of broadcasting.

kyranet:
> I don't think `settingsUpdateEntry.js` needs to be renamed.
> 
> 1. This event is not a discord.js event or any event a normal user would ever be making on their own.
> 2. The only reason a user would make this event is that they have a custom sharding system. Either way this file is named that user would have to transfer this file to disable this form of broadcasting.

- *Some random day in #klasa-support*
- *Guy makes an event called settingsUpdateEntry to update some WeakMaps with brand new data in their sharded bot, and notices the settings don't synchronize between shards.*
- *After a while, we notice it broke because he accidentally overrided the event.*

Yeah, for consistency, all core events should be prefixed, this prevents issues with accidental overrides.

Issue #632, opened 2019-03-18
teams support


bdistin:
### Describe your proposal
This would replace `client.owner: KlasaUser` with `client.owners: Set<KlasaUser>`. This will alter all internal uses of client.owner to give full permissions to everyone in the owners set. This will be filled from the future API endpoint described here <https://github.com/discordapp/discord-api-docs/issues/787>

Edit: going to use a Set for owners instead.

cfanoulis:
this didnt close with #732?

PR #692, opened 2019-05-15
initial teams support

bdistin:
### Description of the PR
This preps the Klasa API for #632 . This finalizes the structure for teams, but without the API endpoint, it won't automatically be filled with the team owners until that is implemented.

This breaks usage of Client#owner and the ownerID client option.

-

### Semver Classification

### DIFF
--- a/src/lib/Client.js
+++ b/src/lib/Client.js
@@ -347,7 +347,7 @@ class KlasaClient extends Discord.Client {
 	 */
 	get owners() {
 		const owners = new Set();
-		for (const owner of this.client.options.owners) {
+		for (const owner of this.options.owners) {
 			const user = this.users.get(owner);
 			if (user) owners.add(user);
 		}
--- a/src/lib/Client.js
+++ b/src/lib/Client.js
@@ -66,7 +66,7 @@ class KlasaClient extends Discord.Client {
 	 * @property {GatewaysOptions} [gateways={}] The options for each built-in gateway
 	 * @property {string} [language='en-US'] The default language Klasa should opt-in for the commands
 	 * @property {boolean} [noPrefixDM=false] Whether the bot should allow prefixless messages in DMs
-	 * @property {string} [ownerID] The discord user id for the user the bot should respect as the owner (gotten from Discord api if not provided)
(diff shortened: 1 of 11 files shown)


src/lib/Client.js
```diff
@@ -340,13 +340,18 @@ class KlasaClient extends Discord.Client {
...
-	 * @type {?KlasaUser}
+	 * The owners for this bot
+	 * @since 0.5.0
+	 * @type {Set<KlasaUser>}
 	 * @readonly
 	 */
-	get owner() {
-		return this.users.get(this.options.ownerID) || null;
+	get owners() {
+		const owners = new Discord.Collection();
```
kyranet:
```suggestion
		const owners = new Set();
```

src/lib/Client.js
```diff
@@ -340,13 +340,18 @@ class KlasaClient extends Discord.Client {
...
+		for (const owner of this.client.options.owners) {
```
tech6hutch: ```this.options.owners```

src/lib/Client.js
```diff
@@ -340,13 +340,18 @@ class KlasaClient extends Discord.Client {
...
 	/**
-	 * The owner for this bot
-	 * @since 0.1.1
-	 * @type {?KlasaUser}
+	 * The owners for this bot
+	 * @since 0.5.0
+	 * @type {Set<KlasaUser>}
 	 * @readonly
 	 */
-	get owner() {
```
tech6hutch: If you wanted to keep some backwards compatibility, you could do
```js
get owner() {
    return this.options.owners.length === 1 ? this.users.get(this.options.owners.values().next().value) : null;
}
```

kyranet: No, that'd make many things harder, overloaded outputs are a pain to maintain.

src/lib/Client.js
```diff
@@ -340,13 +340,18 @@ class KlasaClient extends Discord.Client {
...
-		return this.users.get(this.options.ownerID) || null;
+	get owners() {
+		const owners = new Set();
+		for (const owner of this.options.owners) {
+			const user = this.users.get(owner);
+			if (user) owners.add(user);
```
vladfrangu: Can't we use partial users here?

bdistin: No, that can be a team id currently.

PR #696, opened 2019-05-18
v1.0.0-alpha

### LINKED ISSUE #432: Proposal: Non overriding events

bdistin:
### Description of the PR
This pr joins the half a dozen breaking prs into one big breaking pr

- #632 This breaks usage of Client#owner and the ownerID client option.
- #689 Client is Redundant in Piece constructors since it's on Store the second parameter.
- #682 TextPrompt arbitrary User/Channel (breaks Message#prompt)
- #682 Abort Prompt if retyped Command (breaks custom CommandHandler Monitors)
- #682 Allow flagSupport to be turned off for both TextPrompt and Commands
- #682 Separates argumentErrors from commandError events
- #688 fixes #432 Breaking, renames event files with core Klasa functionality to include a core prefix.

### Semver Classification

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -23,3 +23,4 @@ There are a number of guidelines considered when reviewing Pull Requests to be m
 - Everything should be documented with [jsdocs](http://usejsdoc.org/), whether private or not. __If you see a mistake in the docs, please pr a fix.__
 - Everything should follow OOP paradigms and generally rely on behavior over state where possible. This generally helps methods be predictable, keeps the codebase simple & understandable, reduces code duplication through abstraction, and leads to efficiency and therefore scalability.
 - Everything should follow our ESLint rules as closely as possible, and should pass lint tests even if you must disable a rule for a single line.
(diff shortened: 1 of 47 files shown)


bdistin:
Note: there is a bug in CommandUsage that will require fixing after the Settings branch is merged here. Reminder...

PR #698, opened 2019-05-20
Settings Guides

UnseenFaith:
### Description of the PR
My Guides from the old branch. Don't merge until Kyra makes decision on pluck vs resolve interface.

- New(ish) guides for upcoming Settings update.

### Semver Classification


Skillz4Killz:
Hello 👋 

I was hoping to ask for a bit more information in the `.resolve()` section of the guide. After reading it, I am still quite uncertain of what its purpose is exactly? The description mentions about resolving but what is resolving to mean or do?

`The resolve option allows you to change the behavior of keys that should be resolved when using the resolve method.` This is quite confusing to me.

What I am hoping for is an addition like below which would explain what the resolve does and is useful for.

```js
// This function is used to tell Settings what this data should be stored as
	serialize(value) {
		// Convert Date Object back into milliseconds so it can be stored by the database.
		// Value here is our Date Object from deserialized.
		return value.getTime();
	}```

This is a great example from the guide that helps explain the exact purpose of a function.

**Comment by UnseenFaith** [comment]
I have added more information to the resolve section. Let me know if that makes it worse or better. @Skillz4Killz

**Comment by Skillz4Killz** [comment]
It is Perfect! Really clear and i can easily understand what resolve does now.

P.S. There is a typo in there on line 198

File: guides/Getting Started/Settings.md [hunk 1]
```diff
@@ -0,0 +1,314 @@
...
+
+The resolve option allows you specify whether or not you want your primitive values turned back into their full objects (where possible, ex: Turning User ID => User Object). When using the resolve method, this option will determine whether or not you get back a primitive or not. This option is true by default, and will automatically resolve every single object in array where possible.
+
+An example: You have a roles array that gives certain roles permissions to use more commands. You want to check if a user has any of the roles in the array.
+
+```javascript
+const { Client } = require('klasa');
+
+// Resolve is default by true, here for example purposes
+Client.defaultGuildSceham.add('roles', 'role', { array: true, resolve: true });
```
Skillz4Killz:
```suggestion
Client.defaultGuildSchema.add('roles', 'role', { array: true, resolve: true });
```

guides/Getting Started/Settings.md
```diff
@@ -0,0 +1,314 @@
...
+
+```javascript
+// Get the value from the administrator key from the roles group
+message.guild.settings.get('roles.administrator');
+```
+
+You can use the pluck method to get multiple keys at once. It will return an object keyed by the paths you gave it.
+
+```javascript
+const { prefix, 'roles.administrator': adminRole } = message.guild.settings.pluck('prefix', 'roles.administrator');
```
Skillz4Killz:
```suggestion
const [prefix, adminRole] = message.guild.settings.pluck('prefix', 'roles.administrator');
```

guides/Getting Started/Settings.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+When using nested groups, we separate the groups names by dot:
+
+```javascript
+// Get the value from the administrator key from the roles group
+message.guild.settings.get('roles.administrator');
+```
+
+You can use the pluck method to get multiple keys at once. It will return an object keyed by the paths you gave it.
```
Skillz4Killz:
```suggestion
You can use the pluck method to get multiple keys at once. It will return an array in the same order as the paths you gave it.
```

guides/Getting Started/Settings.md
```diff
@@ -0,0 +1,314 @@
...
+
+```javascript
+const { prefix, 'roles.administrator': adminRole } = message.guild.settings.pluck('prefix', 'roles.administrator');
+```
+
+If you would like resolved values as well, you can use the resolve method instead, it follows the same format as pluck, except that non-resolvable keys will be set to null in the returned object.
```
Skillz4Killz:
```suggestion
If you would like resolved values as well, you can use the resolve method instead, it follows the same format as pluck, except that non-resolvable keys will be set to null in the returned array.
```

guides/Getting Started/Settings.md
```diff
@@ -0,0 +1,314 @@
...
+
+```javascript
+// Can also be awaited and destructured
+// Output: Role Object or null if unresolvable
+message.guild.resolve('roles.administrator').then(resolvedObject => console.log(resolvedObject['roles.administrator']));
```
Skillz4Killz:
```suggestion
message.guild.resolve('roles.administrator').then(resolvedArray => console.log(resolvedArray[0]));
```

PR #700, opened 2019-05-21
Convert .pluck and .resolve To Return Arrays

Skillz4Killz:
### Description of the PR
This PR changes `.pluck()` and `.resolve()` to return arrays instead of objects.

Instead of returning an object to deconstruct, returning an array in the order of the arguments provided which can be deconstructed is better.

```js
const { prefix, 'roles.administrator': adminRole } = message.guild.settings.pluck('prefix', 'roles.administrator');
```

to

```js
const [prefix, adminRole] = message.guild.settings.pluck('prefix', 'roles.administrator')
```

No need to type the path to the key twice.

- Change .pluck and .resolve return type to array

### Semver Classification

### DIFF
--- a/src/lib/settings/SettingsFolder.js
+++ b/src/lib/settings/SettingsFolder.js
@@ -120,15 +120,13 @@ class SettingsFolder extends Map {
  	 * @param  {...string} paths The paths to resolve
  	 * @returns {Promise<any[]>}
  	 */
-	async resolve(...paths) {
+	resolve(...paths) {
 		const guild = resolveGuild(this.base.gateway.client, this.base.target);
 		const language = guild ? guild.language : this.base.gateway.client.languages.default;
-		const promises = [];
-		const values = await Promise.all(paths.map(path => {
+		return Promise.all(paths.map(path => {
 			const entry = this.schema.get(this.relative(path));
 			return entry.resolve(this, language, guild).then(res => res)
 		}));
-		return values;
 	}
 
 	/**
--- a/src/lib/settings/SettingsFolder.js
+++ b/src/lib/settings/SettingsFolder.js
@@ -118,7 +118,7 @@ class SettingsFolder extends Map {
  	 * Resolves paths into their full objects or values depending on the current set value
  	 * @since 0.5.0
  	 * @param  {...string} paths The paths to resolve
- 	 * @returns {Promise<any[]>}
+ 	 * @returns {Promise<Array<*>>}
  	 */
 	resolve(...paths) {
 		const guild = resolveGuild(this.base.gateway.client, this.base.target);
--- a/src/lib/settings/SettingsFolder.js
+++ b/src/lib/settings/SettingsFolder.js
@@ -120,17 +120,17 @@ class SettingsFolder extends Map {
  	 * Resolves paths into their full objects or values depending on the current set value
  	 * @since 0.5.0
  	 * @param  {...string} paths The paths to resolve
- 	 * @returns {*}
+ 	 * @returns {Promise<any[]>}
  	 */
 	async resolve(...paths) {
 		const guild = resolveGuild(this.base.gateway.client, this.base.target);
 		const language = guild ? guild.language : this.base.gateway.client.languages.default;
 		const promises = [];
-		for (const path of paths) {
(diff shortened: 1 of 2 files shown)


src/lib/settings/SettingsFolder.js
```diff
@@ -120,17 +118,17 @@ class SettingsFolder extends Map {
...
  	 * @param  {...string} paths The paths to resolve
- 	 * @returns {*}
+ 	 * @returns {Promise<any[]>}
  	 */
 	async resolve(...paths) {
 		const guild = resolveGuild(this.base.gateway.client, this.base.target);
 		const language = guild ? guild.language : this.base.gateway.client.languages.default;
 		const promises = [];
-		for (const path of paths) {
+		const values = await Promise.all(paths.map(path => {
```
MrJacz: Just return the `Promise.all` and remove async from the method

src/lib/settings/SettingsFolder.js
```diff
@@ -120,17 +118,17 @@ class SettingsFolder extends Map {
  	 * Resolves paths into their full objects or values depending on the current set value
  	 * @since 0.5.0
  	 * @param  {...string} paths The paths to resolve
- 	 * @returns {*}
+ 	 * @returns {Promise<any[]>}
```
MrJacz: using `Promise` in jsdoc when the method is async is redunant due to docsgen (the tool that builds the docs json file) does this for us. but since we're going to remove async from the function this can stay

src/lib/settings/SettingsFolder.js
```diff
@@ -120,17 +118,15 @@ class SettingsFolder extends Map {
...
 		const guild = resolveGuild(this.base.gateway.client, this.base.target);
 		const language = guild ? guild.language : this.base.gateway.client.languages.default;
-		const promises = [];
-		for (const path of paths) {
+		return Promise.all(paths.map(path => {
 			const entry = this.schema.get(this.relative(path));
-			promises.push(entry.resolve(this, language, guild).then(res => ({ [entry.key]: res })));
-		}
-		return Object.assign({}, ...await Promise.all(promises));
+			return entry.resolve(this, language, guild).then(res => res);
```
vladfrangu: Any reason for the `then(res => res)`?

src/lib/settings/SettingsFolder.js
```diff
@@ -88,18 +88,16 @@ class SettingsFolder extends Map {
...
 	 */
 	pluck(...paths) {
-		const object = {};
-		for (const path of paths) {
+		return paths.map(path => {
 			const value = this.get(path);
-			if (typeof value !== 'undefined') object[path] = value instanceof SettingsFolder ? value.toJSON() : value;
-		}
-		return object;
+			return typeof value !== 'undefined' ? value instanceof SettingsFolder ? value.toJSON() : value : null;
```
vladfrangu: Defaulting values that don't exist to null is a bad idea... If I pluck `myValueThatTotallyDoesntExistPerMySchema` I expect it to be undefined, not null... And thats what the old code did

src/lib/settings/SettingsFolder.js
```diff
@@ -88,18 +88,16 @@ class SettingsFolder extends Map {
 	 * Plucks out one or more attributes from either an object or a sequence of objects
 	 * @since 0.5.0
 	 * @param  {...string} paths The paths to take
-	 * @returns {Object<string,*>}
+	 * @returns {Array<any>}
```
vladfrangu:
```suggestion
	 * @returns {Array<*>}
```

src/lib/settings/SettingsFolder.js
```diff
@@ -120,17 +118,15 @@ class SettingsFolder extends Map {
  	 * Resolves paths into their full objects or values depending on the current set value
  	 * @since 0.5.0
  	 * @param  {...string} paths The paths to resolve
- 	 * @returns {*}
+ 	 * @returns {Promise<any[]>}
```
vladfrangu:
```suggestion
 	 * @returns {Promise<Array<*>>}
```

Skillz4Killz:
Requested Changes Done

PR #721, opened 2019-06-07
typings: Mark Task#run's return type as unknown

Skillz4Killz:
### Description of the PR
This PR allows you to return anything from a task. It is really annoying to be forced to return void in tasks. In every one of my files when I want to cancel something I always do `return null` but in the tasks, I am being forced to do `return`. Forcing a return void destroys consistency across my bot.

Also, the fact that I can't just end the task like
```js
return this.client.emit('log', 'x task ran successfully')
```

But instead, have to do
```js
this.client.emit('log', 'x task ran successfully')
return
```

```js
Property 'run' in type 'default' is not assignable to the same property in base type 'Task'.
  Type '() => Promise<boolean | void>' is not assignable to type '(data: any) => Promise<void>'.
    Type 'Promise<boolean | void>' is not assignable to type 'Promise<void>'.
      Type 'boolean | void' is not assignable to type 'void'.
        Type 'false' is not assignable to type 'void'.ts(2416)
```

- Promise<void> => Promise<any>

### Semver Classification

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -517,7 +517,7 @@ declare module 'klasa' {
 
 	export abstract class Task extends Piece {
 		public constructor(client: KlasaClient, store: TaskStore, file: string[], directory: string, options?: TaskOptions);
-		public abstract run(data: any): Promise<unknown>;
+		public abstract run(data: any): unknown;
 		public toJSON(): PieceTaskJSON;
 	}
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -517,7 +517,7 @@ declare module 'klasa' {
 
 	export abstract class Task extends Piece {
 		public constructor(client: KlasaClient, store: TaskStore, file: string[], directory: string, options?: TaskOptions);
-		public abstract run(data: any): Promise<void>;
+		public abstract run(data: any): Promise<any>;
 		public toJSON(): PieceTaskJSON;
 	}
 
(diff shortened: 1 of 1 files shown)


tech6hutch:
What's the use case for returning something from Task#run? Running tasks manually?

Skillz4Killz:
```js
if (somecheck) {
  this.client.emit(`log`, `explain why it was cancelled`)
  return
}
```

As opposed to

```js
if (somecheck) return this.client.emit(`log`, `explian why it was cancelled`)

kyranet:
In that case, the return type should be `unknown`, as it'll act like a rubbish bin type

Skillz4Killz:
Do you mean it should return Promise <unknown>

tech6hutch:
Promise<unknown>

tech6hutch:
Whoa, GitHub eats the part in less-than greater-than signs. `Promise<unknown>`

kyranet:
`Promise<unknown>` and `unknown` differ that in one the promise is required, and the other can be a promise, can be a string, can be a Tesla Roadster, can be anything, since it's unknown.

It's a type implemented in TypeScript 3.2.0 that can be casted to everything.

Skillz4Killz:
@bdistin was this closed accidentally without merging or did you have a good reason to close this? I thought i gave a pretty good use case for this change.

bdistin:
the change is already in alpha

PR #724, opened 2019-06-12
prevent duplicate aliases (master branch)

BannerBomb:
### Description of the PR
Added a check to the autoAliases method to prevent aliases from being added if it is already present in the aliases array. Possibly added by the user manually or if you have aliases like `this-is-an-example`,`this-isan-example`,`thisisan-example` which would normally end up changing them all to `thisisanexample` and adding them to the aliases array.

-

### Semver Classification


bdistin:
I addressed why this was denied in chat

PR #731, opened 2019-06-15
prevent duplicate aliases (settings banch)

BannerBomb:
### Description of the PR
Added a check to the autoAliases method to prevent aliases from being added if it is already present in the aliases array. Possibly added by the user manually or if you have aliases like `this-is-an-example`, `this-isan-example`, `thisisan-example` which would normally end up changing them all to `thisisanexample` and adding them to the aliases array.

-

### Semver Classification

### DIFF
--- a/src/lib/structures/Command.js
+++ b/src/lib/structures/Command.js
@@ -56,8 +56,8 @@ class Command extends AliasPiece {
 		this.name = this.name.toLowerCase();
 
 		if (options.autoAliases) {
-			if (this.name.includes('-')) this.aliases.push(this.name.replace(/-/g, ''));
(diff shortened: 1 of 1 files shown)


bdistin:
this isn't necessary, as the end result is still a collection mapping aliases to commands, which you can't have duplicate keys.

BannerBomb:
I know the aliases only have 1 alias mapped to a command, but it affects if you have a `help` embed for a command that lists all the command's aliases. Like this 
image

This happened because `autoAliases` automatically added the aliases while I also had the aliases already defined in the alias array for that command.

bdistin:
If you don't like that behavior, then don't do it. There is no need to patch for that behavior directly in klasa.

PR #733, opened 2019-06-21
misc: Freeze all objects from Settings

kyranet:
### Description of the PR

TypeScript's hints weren't enough, so let's get the `readonly` into the real world.

- Freeze absolutely all objects cached from SG.

### Semver Classification

### DIFF
--- a/src/lib/util/util.js
+++ b/src/lib/util/util.js
@@ -142,6 +142,26 @@ class Util {
 	}
 
 	/**
+	 * Deep freeze a value
+	 * @since 0.5.0
+	 * @param {*} source The object to freeze
+	 * @returns {Readonly<*>}
+	 */
+	static deepFreeze(source) {
+		if (typeof source === 'object' && source !== null) {
+			Object.freeze(source);
+			if (Array.isArray(source)) {
+				for (const value of source) Util.deepFreeze(value);
+			} else {
+				for (const value of Object.values(source)) {
+					Util.deepFreeze(value);
+				}
+			}
+		}
+		return source;
+	}
+
+	/**
 	 * Verify if the input is a function.
 	 * @since 0.5.0
 	 * @param {*} input The function to verify
--- a/src/lib/util/util.js
+++ b/src/lib/util/util.js
@@ -152,10 +152,12 @@ class Util {
 			Object.freeze(source);
 			if (Array.isArray(source)) {
 				for (const value of source) Util.deepFreeze(value);
+			} else if (source instanceof Map) {
+				for (const value of source.values()) Util.deepFreeze(value);
+			} else if (source instanceof Set) {
+				for (const value of source) Util.deepFreeze(value);
 			} else {
-				for (const value of Object.values(source)) {
-					Util.deepFreeze(value);
-				}
+				for (const value of Object.values(source)) Util.deepFreeze(value);
 			}
 		}
 		return source;
--- a/src/lib/settings/SettingsFolder.js
+++ b/src/lib/settings/SettingsFolder.js
@@ -435,7 +435,7 @@ class SettingsFolder extends Map {
 
 			// Patch recursively if the key is a folder, set otherwise
 			if (subkey instanceof SettingsFolder) subkey._patch(value);
-			else super.set(key, value);
+			else super.set(key, Object.freeze(value));
 		}
 	}
 
--- a/src/lib/settings/SettingsFolder.js
+++ b/src/lib/settings/SettingsFolder.js
@@ -1,4 +1,4 @@
-const { isObject, objectToTuples, arraysStrictEquals, toTitleCase, mergeObjects, makeObject, resolveGuild } = require('../util/util');
+const { isObject, objectToTuples, arraysStrictEquals, toTitleCase, mergeObjects, makeObject, resolveGuild, deepFreeze } = require('../util/util');
 const Type = require('../util/Type');
 
 /**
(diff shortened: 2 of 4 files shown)


tech6hutch:
I think this is overall a very good change 👍 

How much of a performance impact does freezing everything in the cache have? It would be nice to have an option to turn this off, for experienced users.

kyranet:
> How much of a performance impact does freezing everything in the cache have? It would be nice to have an option to turn this off, for experienced users.

In a large bot, how long does it take for V8 to do `object.#flags |= FROZEN | SEALED;` (with obviously all the other spec stuff)?

I don't know, you can benchmark it, but I doubt it's that heavy, in my bot, only a small portion of the settings are different from defaults, and those are the ones that get individually frozen.

Initialization uses SettingsFolder#init, which doesn't use Object.freeze, since defaults are frozen with this change.

src/lib/util/util.js
```diff
@@ -141,6 +141,26 @@ class Util {
...
+	 * @since 0.5.0
+	 * @param {*} source The object to freeze
+	 * @returns {Readonly<*>}
+	 */
+	static deepFreeze(source) {
+		if (typeof source === 'object' && source !== null) {
+			Object.freeze(source);
+			if (Array.isArray(source)) {
+				for (const value of source) Util.deepFreeze(value);
+			} else {
```
tech6hutch: This function will not work on maps or sets. Their keys/values won't be frozen.

kyranet: Fixed in 02d6353

src/lib/util/util.js
```diff
@@ -152,10 +152,12 @@ class Util {
 			Object.freeze(source);
 			if (Array.isArray(source)) {
 				for (const value of source) Util.deepFreeze(value);
+			} else if (source instanceof Map) {
+				for (const value of source.values()) Util.deepFreeze(value);
```
tech6hutch: This freezes map values, but not keys. Is it unimportant that they be frozen?

kyranet: They're always primitives (strings, booleans, numbers...), they cannot be frozen.

tech6hutch: Do you mean in the context of SG, map keys will always be primitives? Because they can be objects in JS.

Skillz4Killz:
I did some research on this and it appears that Object.freeze is really performance heavy. There is a 67% drop in performance. https://jsperf.com/performance-frozen-object/74

I think Ao touched on this point once in Discord that its the developers own responsibility not to edit their own cached settings. Normal/proper use of Klasa SG would never need this change. This could add a heavy overload to performance to every bot for something that is fixed simply with the `readonly`. If someone breaks their settings even after the readonly then it’s really their own mistake. Every bot should not have to suffer heavily in terms of performance to prevent this behavior.

If this is added, there should be an option to opt in or opt out.

tech6hutch:
> I did some research on this and it appears that Object.freeze is really performance heavy. There is a 67% drop in performance. https://jsperf.com/performance-frozen-object/74

Thank you for doing the research.

> I think Ao touched on this point once in Discord that its the developers own responsibility not to edit their own cached settings. Normal/proper use of Klasa SG would never need this change.

If we're going to take that angle, then we really need to hammer the point home in the docs and everywhere. I think we should just avoid having this footgun, however.

> This could add a heavy overload to performance to every bot for something that is fixed simply with the `readonly`. If someone breaks their settings even after the readonly then it’s really their own mistake. Every bot should not have to suffer heavily in terms of performance to prevent this behavior.

That only works in TypeScript bots. This is a JavaScript library; TS support is a bonus.

> If this is added, there should be an option to opt in or opt out.

I wouldn't have a problem with this. I really think the default should be freezing, however.

vladfrangu:
> I did some research on this and it appears that Object.freeze is really performance heavy. There is a 67% drop in performance. jsperf.com/performance-frozen-object/74

The test you linked ran `Object.freeze` on every iteration, which, of course, slowed it down a LOT. If you freeze the object beforehand, THEN run the test, as done here: https://jsperf.com/frozen-vs-non-frozen-for-of-and-for-in

You'll get the following results

image

The difference is negligible. The cost for freezing happens only once, at constructing the object, thereby in the end, performance *should* be the same

kyranet:
> I did some research on this and it appears that Object.freeze is really performance heavy. There is a 67% drop in performance. https://jsperf.com/performance-frozen-object/74

Can run 34M times a second, that's fine, considering almost nobody uses the `any` type or arrays in user schemas, and a properly-made but also large bot would store ~10-20 keys with that type, I don't think I'll see a second of slowdown, and this is really helpful to tell you not to touch your cache accidentally, thus preventing a lot of bugs.

What is frozen? The default values from all SchemaPieces, and any non-primitive value that is patched (modified values that are stored from the db), in my bot for example, deepFreeze would only run about ~3500 times at startup, for 715 guilds and 356k users (counting the recursion too). _And I can run it 1000 times more_.

Tylertron1998:
> If this is added, there should be an option to opt in or opt out.
> I wouldn't have a problem with this. I really think the default should be freezing, however.

I agree here - while this is a really good idea for people who don't quite understand the workings of SettingsGateway, or just generally as a hint to not modify the cache directly, this does have a performance impact compared to not deepFreezing - sure, it may be pretty small, but I can bet for some very big bots who are constantly updating/changing settings, that performance does matter.

kyranet:
> but I can bet for some very big bots who are constantly updating/changing settings, that performance does matter.

We're talking about a decrease of performance in the factor of a dozen nanoseconds for large updates, and the scenario of an array/any value being updated all time is extremely unlikely.

kyranet:
This won't be needed with #752.

PR #743, opened 2019-07-04
Fix cooldown response

Skillz4Killz:
### Description of the PR
This PR fixes the response of the cooldown inhibitor when the cooldown level is for a guild. At the moment, it always responds saying **you** as opposed to someone even if someone else used that command. It also changes the word `just` => `already`. In case, commands that have long cooldowns like 24 hours, this can be extremely confusing making users think something is wrong because they did not JUST use it.

- Fix inhibitor cooldown response

### Semver Classification

### DIFF
--- a/src/inhibitors/cooldown.js
+++ b/src/inhibitors/cooldown.js
@@ -11,7 +11,7 @@ module.exports = class extends Inhibitor {
 
 		const existing = command.cooldowns.get(message.levelID);
 
-		if (existing && existing.limited) throw message.language.get('INHIBITOR_COOLDOWN', Math.ceil(existing.remainingTime / 1000), command.cooldownLevel === 'guild');
+		if (existing && existing.limited) throw message.language.get('INHIBITOR_COOLDOWN', Math.ceil(existing.remainingTime / 1000), command.cooldownLevel !== 'author');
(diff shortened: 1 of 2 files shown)


tech6hutch:
If someone has a long cooldown, they could change the wording themselves imo. But good PR.

src/inhibitors/cooldown.js
```diff
@@ -11,7 +11,7 @@ module.exports = class extends Inhibitor {
 
 		const existing = command.cooldowns.get(message.levelID);
 
-		if (existing && existing.limited) throw message.language.get('INHIBITOR_COOLDOWN', Math.ceil(existing.remainingTime / 1000));
+		if (existing && existing.limited) throw message.language.get('INHIBITOR_COOLDOWN', Math.ceil(existing.remainingTime / 1000), command.cooldownLevel === 'guild');
```
kyranet: This should be different to author, instead of equal to guild, cooldownLevel can also be configured to channel.

```suggestion
		if (existing && existing.limited) throw message.language.get('INHIBITOR_COOLDOWN', Math.ceil(existing.remainingTime / 1000), command.cooldownLevel !== 'author');
```

PR #752, opened 2019-07-11
Settings Map/Set Implementation

UnseenFaith:
### Description of the PR
This PR adds the ability to use Map/Set within Settings, along with array changes to make the entire interface more consistent. (This is WIP and not ready to be consumed in any way, shape, or form)

## Array
A new class called SettingsArray that all arrays will use. It does not extend Arrays as to not allow end-users to mutate the returned array. This class will hold several methods equivalent to normal array operations, the only difference being you won't be able to mutate the Array in cache (without willingly trying to).

#### Methods:
- [x] get - equivalent to Map/Set.get(); Takes a value or no value and returns clonedArray[index] or clonedArray
- [x] includes - equivalent to Array includes
- [x] update - new update method to update Arrays, SettingsFolder#update will route to this
- [ ] reset - new reset method to reset Arrays, SettingsFolder#reset will route to this
- [ ] forEach - equivalent to forEach except with a cloned array
- [x] Symbol.iterator - equivalent to Array iterator
- [ ] find - equivalent to Array find

> If you think other methods should be added, feel free to comment. I'm not trying to fill out the entire array prototype but I don't mind having helpful methods added back.

#### How to use the update function:
Assuming we have the array `[1,2,3,4,5]`, and all methods without an action are defaulted to 'auto'
Single Value:
```javascript
update(6); // [1,2,3,4,5,6]
```
Multiple Values:
```javascript
update([6, 7, 8]); // [1,2,3,4,5,6,7,8]
```
Single value by index:
```javascript
update([[2, 6]]); // [1,2,6,4,5]
```
Multiple values by index:
```javascript
update([[0, 6], [4, 7]]); // [6,2,3,4,7]
```
> Indexing bypasses all actions passed, and you cannot combine indexing with non-indexing.

I'll add the specifics here when the implementation is finished.

### Semver Classification

### DIFF
--- a/src/lib/settings/group/GroupBase.js
+++ b/src/lib/settings/group/GroupBase.js
@@ -1,0 +1,48 @@
+class GroupBase {
+
+	constructor(entry) {
+		Object.defineProperty(this, 'base', { value: null, writable: true });
+
+		// The Entry this SettingArray refers to
+		Object.defineProperty(this, 'entry', { value: entry });
+
+		this.data = entry.default;
+	}
+
+	get gateway() {
+		return this.base.gateway;
+	}
+
+	async _save(results) {
+		const status = this.base.existenceStatus;
+		if (status === null) throw new Error('Cannot update out of sync.');
+
+		// Update DB
+
+		this._patch(results[0].value);
+	}
+
+	*keys() {
+		yield* this.data.keys();
+	}
+
+	*values() {
+		yield* this.data.values();
+	}
+
+	*entries() {
+		yield* this.data.entries();
+	}
+
+	*[Symbol.iterator]() {
+		yield* this.data[Symbol.iterator]();
+	}
+
+	// This should work fine in all 3 cases
+	toJSON() {
+		return [...this[Symbol.iterator]()];
+	}
+
+}
+
+module.exports = GroupBase;
--- a/src/lib/settings/group/GroupBase.js
+++ b/src/lib/settings/group/GroupBase.js
@@ -1,18 +1,88 @@
+/**
+ * The base class for group structs
+ */
 class GroupBase {
 
+	/**
+	 * @since 0.5.0
+	 * @param {SchemaEntry} entry The schema entry that manages this instance
+	 */
 	constructor(entry) {
+		/**
+		 * The base that owns this instance's data
+		 * @since 0.5.0
+		 * @name GroupBase#base
+		 * @type {?Settings}
+		 */
 		Object.defineProperty(this, 'base', { value: null, writable: true });
 
-		// The Entry this SettingArray refers to
+		/**
+		 * The schema entry that manage this instance's metadata
+		 * @since 0.5.0
+		 * @name GroupBase#entry
+		 * @type {SchemaEntry}
+		 * @readonly
+		 */
 		Object.defineProperty(this, 'entry', { value: entry });
 
+		/**
+		 * The data this instance holds, it is initially initialized as the schema entry's default
+		 * @since 0.5.0
+		 * @type {*}
+		 */
 		this.data = entry.default;
 	}
 
+	/**
+	 * The gateway that manages this instance's table
+	 * @since 0.5.0
+	 * @type {Gateway}
+	 * @readonly
+	 */
 	get gateway() {
 		return this.base.gateway;
 	}
 
+	/**
+	 * The get method to be overwritten
+	 * @since 0.5.0
+	 * @param {*} key The key to be retrieved
+	 * @abstract
+	 */
+	get() {
+		// Defined in extension Classes
+		throw new Error(`[GroupBase] Missing method 'update' of ${this.constructor.name}`);
+	}
+
+	/**
+	 * The update method to be overwritten
+	 * @since 0.5.0
+	 * @param {...*} param The parameters to be passed
+	 * @abstract
+	 */
+	update() {
+		// Defined in extension Classes
+		throw new Error(`[GroupBase] Missing method 'update' of ${this.constructor.name}`);
+	}
+
+	/**
+	 * The _patch method to be overwritten
+	 * @since 0.5.0
+	 * @param {*} param The value to be updated
+	 * @protected
+	 * @abstract
+	 */
+	_patch() {
+		// Defined in extension Classes
+		throw new Error(`[GroupBase] Missing method 'update' of ${this.constructor.name}`);
+	}
+
+	/**
+	 * Save the data to the database and patch the data.
+	 * @since 0.5.0
+	 * @param {Array<any[]>} results The data to save
+	 * @protected
+	 */
 	async _save(results) {
 		const status = this.base.existenceStatus;
 		if (status === null) throw new Error('Cannot update out of sync.');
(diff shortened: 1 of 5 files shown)


bdistin:
This pull request **introduces 2 alerts** when merging eb09cfd41d478dc1e20cc57bdca76316e8fe2fa8 into 412335579d4ed365d67bbcfff66c2f6a505fd2b4 - [view on LGTM.com](https://lgtm.com/projects/g/dirigeants/klasa/rev/pr-43a61d34d3b734549fa8a6c92eb73e93bab5bf1f)

**new alerts:**

* 2 for Unused variable, import, function or class

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

*Comment posted by [LGTM.com](https://lgtm.com)*

src/lib/settings/group/SettingsArray.js
```diff
@@ -0,0 +1,127 @@
...
+
+		const clone = this.get();
+
+		// Errors are given back in the order that values were sent in
+		const errors = [];
+
+		// This value has an index paired with it
+		if (indexing) {
+			for (const val of values) {
+				let index = val[0];
```
vladfrangu: Could use some destructuring, use it at the bottom

```suggestion
				let [index, actualValue] = val;
```

src/lib/settings/group/SettingsArray.js
```diff
@@ -0,0 +1,127 @@
...
+
+		this.data = entry.default;
+	}
+
+	get gateway() {
+		return this.base.gateway;
+	}
+
+	get(index) {
+		if (index !== undefined && index !== null) {
```
vladfrangu: nit: Could probably use `!= undefined` and disable the eslint rule on the line

src/lib/settings/group/SettingsArray.js
```diff
@@ -0,0 +1,127 @@
...
+		// This value has an index paired with it
+		if (indexing) {
+			for (const val of values) {
+				let index = val[0];
+				if (clone.length === 0 && index > 0) {
+					errors.push({ input: val, message: 'The current array is empty. The index must start at 0.' });
+				} else if (index < 0 || index > clone.length + 1) {
+					errors.push({ input: val, message: `The index ${index} is bigger than the current array. It must be a value in the range of 0..${clone.length + 1}.` });
+				}
+				clone[index] = val[1];
```
vladfrangu: This will always run, even with the errors.. Is that intended?

src/lib/settings/group/SettingsArray.js
```diff
@@ -0,0 +1,127 @@
...
+			}
+		} else if (action === 'auto') {
+			for (const val of values) {
+				const index = clone.indexOf(val);
+				if (index === -1) clone.push(val);
+				else clone.splice(index, 1)
+			}
+		} else if (action === 'add') {
+			for (const val of values) {
+				if (clone.includes(val)) errors.push({ input: val, message: `The value ${val} for the key ${entry.path} already exists.` });
```
vladfrangu:
```suggestion
				if (clone.includes(val)) errors.push({ input: val, message: `The value "${val}" for the key "${entry.path}" already exists.` });
```

src/lib/settings/group/SettingsArray.js
```diff
@@ -0,0 +1,127 @@
...
+				else clone.push(val);
+			}
+		} else if (action === 'remove') {
+			for (const val of values) {
+				const index = clone.indexOf(val);
+				if (index === -1) errors.push({ input: val, message: `The value ${val} for the key ${entry.path} does not exist.` });
```
vladfrangu:
```suggestion
				if (index === -1) errors.push({ input: val, message: `The value "${val}" for the key "${entry.path}" does not exist.` });
```

src/lib/settings/group/SettingsArray.js
```diff
@@ -0,0 +1,127 @@
...
+				else clone.splice(index, 1);
+			}
+		} else {
+			throw `The ${action} array action is not a valid SettingsUpdateArrayAction.`;
```
vladfrangu: Also, this doesn't handle upper case... Should we handle that too? probably not...

```suggestion
			throw `The "${action}" array action is not a valid SettingsUpdateArrayAction.`;
```

PyroTechniac:
Just a recommendation, but the Array#find() method would be nice for large arrays of objects to find one with a specific key, example use case would be finding a specific mod log or a user profile

Skillz4Killz:
Hello 👋 

After reading the PR, I still don’t really understand the benefits of this change. As far as I am aware we were always able to update arrays and add/remove/edit multiple items. The only difference is that now you can add/remove a specific index? 

Also how does a user use this class? 

How does this PR affect the following line below? I mean it uses an Array so does that get affected anyhow?
```js
guildsSchema
  .add(‘roleIDs’, ‘role’, { array: true })
```

UnseenFaith:
@PyroTechniac Will add that to the list.
@Skillz4Killz Beyond the obvious benefits of the map/set implementations allowing you to use said objects inside of Settings, the array specifically should make it more logical when working with arrays, and will also make it not as easy to mutate the array returned from the settings (causing a desync between the cache and your database). There's no immediate changes right now for that specific code line, but I can definitely say that the array option will probably be removed in favor of a different option that will encase arrays/sets/maps (and by extension, in the future, anything that extends those three options)

The idea of the array implementation is to encase it within an object (SettingsArray) so that when people access it they aren't going to desync their settings, alongside providing methods to work with Arrays to make it just as easy as before. Extending Arrays is not an option because we cannot overload operators in Javascript (we can't change how property accessing works ) and we also wouldn't be able to use default to their complete benefit to keep memory low.

src/lib/settings/group/SettingsArray.js
vladfrangu: In fact, this should use util.isNumber

src/lib/settings/group/SettingsArray.js
```diff
@@ -0,0 +1,155 @@
...
+		}
+		return this.data.slice();
+	}
+
+	includes(value) {
+		return this.data.includes(value);
+	}
+
+	find(...args) {
+		return this.data.slice().find(...args);
```
vladfrangu: Any reason for the slice? `find` doesn't alter the array in place.. 
Don't we want to deepClone the result? Then I could see the use for slice, but in my opinion it'd be best to deepClone the result of find

src/lib/settings/group/SettingsArray.js
```diff
@@ -0,0 +1,155 @@
...
+		const errors = [];
+
+		// This value has an index paired with it
+		if (indexing) {
+			for (const val of values) {
+				const [index, value] = val;
+				if (clone.length === 0 && index > 0) {
+					errors.push({ input: val, message: 'The current array is empty. The index must start at 0.' });
+				} else if (index < 0 || index > clone.length + 1) {
+					errors.push({ input: val, message: `The index ${index} is bigger than the current array. It must be a value in the range of 0..${clone.length + 1}.` });
```
vladfrangu: That `clone.length + 1` looks wrong, considering the check is `index > clone.length + 1`

src/lib/settings/group/SettingsArray.js
```diff
@@ -0,0 +1,155 @@
...
+	*keys() {
+		yield* this.data.keys();
+	}
+
+	*values() {
+		yield* this.data.values();
+	}
+
+	*entries() {
+		yield* this.data.values();
```
vladfrangu:
```suggestion
		yield* this.data.entries();
```

ImUrX:
What about implementing .some()?

src/lib/settings/group/GroupBase.js
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * The get method to be overwritten
+	 * @since 0.5.0
+	 * @param {*} key The key to be retrieved
+	 * @abstract
+	 */
+	get() {
+		// Defined in extension Classes
+		throw new Error(`[GroupBase] Missing method 'update' of ${this.constructor.name}`);
```
tjrgg:
```suggestion
		throw new Error(`[GroupBase] Missing method 'get' of ${this.constructor.name}`);
```

src/lib/settings/group/GroupBase.js
```diff
@@ -0,0 +1,147 @@
...
+	/**
+	 * The _patch method to be overwritten
+	 * @since 0.5.0
+	 * @param {*} param The value to be updated
+	 * @protected
+	 * @abstract
+	 */
+	_patch() {
+		// Defined in extension Classes
+		throw new Error(`[GroupBase] Missing method 'update' of ${this.constructor.name}`);
```
tjrgg:
```suggestion
		throw new Error(`[GroupBase] Missing method '_patch' of ${this.constructor.name}`);
```

UnseenFaith:
Last few settings PRs have made this one obsolete, and map/set isn't a high priority (similar functionality already exists, arrayAction: auto and tuples)

PR #754, opened 2019-07-11
Update index.d.ts

Skillz4Killz:
### Description of the PR
This PR will allow commands to be return with Message types. For example, 
```js
return message.channel.send(…)
```

- Allow Message to be returned in commands

### Semver Classification


tech6hutch:
Why not remove `KlasaMessage | KlasaMessage[]`, then? They're Messages.

Skillz4Killz:
@tech6hutch Would that still allow `return message.send()` or message.sendLocale etc…

PR #762, opened 2019-07-19
perf(Timestamp): Increased performance and move patterns to maps

kyranet:
### Description of the PR

There were two branches V8 marked as megamorphic:

https://github.com/dirigeants/klasa/blob/3360292f349cc791063a6b4dcb709080c486a1d8/src/lib/util/Timestamp.js#L109

At `klasa/src/lib/util/Timestamp.js:109:75`

Old State | New State | Key | Map
-- | -- | -- | --
unintialized | premonomorphic | YYYY | 0x211d961a39e9
premonomorphic | monomorphic | MM | 0x211d961a39e9
monomorphic | megamorphic | DD | 0x211d961a39e9

---

https://github.com/dirigeants/klasa/blob/3360292f349cc791063a6b4dcb709080c486a1d8/src/lib/util/Timestamp.js#L127

At `klasa/src/lib/util/Timestamp.js:127:69`

Old State | New State | Key | Map
-- | -- | -- | --
unintialized | premonomorphic | Y | 0x211d96198f11
premonomorphic | monomorphic | Y | 0x211d96198f11
monomorphic | megamorphic | M | 0x211d96198f11

---

This also implements a breaking change, moving all public and exported `Timestamp[token]` to a private and non-exposed `Map`, following `Duration`'s design:

https://github.com/dirigeants/klasa/blob/3360292f349cc791063a6b4dcb709080c486a1d8/src/lib/util/Duration.js#L1-L47

Also adds `Timestamp.mapTokens` and documented `timezoneOffset`.

- Added `Timestamp.mapTokens` and documented `Timestamp.timezoneOffset`.
- Removed all `Timestamp[token]`.

### Semver Classification

### DIFF
--- a/src/lib/util/Timestamp.js
+++ b/src/lib/util/Timestamp.js
@@ -72,8 +72,6 @@ const tokens = new Map([
 ]);
 /* eslint-enable max-len */
 
-const tokenRepeatingCounts = new Map(Object.entries(TOKENS));
-
 /**
  * Klasa's Timestamp class, parses the pattern once, displays the desired Date or UNIX timestamp with the selected pattern.
  */
@@ -196,7 +194,7 @@ class Timestamp {
 		for (let i = 0; i < pattern.length; i++) {
 			let current = '';
 			const currentChar = pattern[i];
-			const tokenMax = tokenRepeatingCounts.get(currentChar);
+			const tokenMax = TOKENS.get(currentChar);
 			if (typeof tokenMax === 'number') {
 				current += currentChar;
 				while (pattern[i + 1] === currentChar && current.length < tokenMax) current += pattern[++i];
@@ -207,7 +205,7 @@ class Timestamp {
 				template.push({ type: 'literal', content: current });
 			} else {
 				current += currentChar;
-				while (i + 1 < pattern.length && !tokenRepeatingCounts.has(pattern[i + 1]) && pattern[i + 1] !== '[') current += pattern[++i];
+				while (i + 1 < pattern.length && !TOKENS.has(pattern[i + 1]) && pattern[i + 1] !== '[') current += pattern[++i];
 				template.push({ type: 'literal', content: current });
 			}
 		}
--- a/src/lib/util/Timestamp.js
+++ b/src/lib/util/Timestamp.js
@@ -1,116 +1,76 @@
 const { TIME: { SECOND, MINUTE, DAY, DAYS, MONTHS, TIMESTAMP: { TOKENS } } } = require('./constants');
 
-/* eslint-disable id-length, new-cap */
-
-// Dates
-
-const Y = time => String(time.getFullYear()).slice(0, 2);
-const YYY = time => String(time.getFullYear());
-const Q = time => String((time.getMonth() + 1) / 3);
-const M = time => String(time.getMonth() + 1);
-const MM = time => M(time).padStart(2, '0');
-const MMM = time => MONTHS[time.getMonth()];
-const D = time => String(time.getDate());
-const DD = time => D(time).padStart(2, '0');
-const DDD = time => {
-	const start = new Date(time.getFullYear(), 0, 0);
-	const diff = ((time.getMilliseconds() - start.getMilliseconds()) + (start.getTimezoneOffset() - time.getTimezoneOffset())) * MINUTE;
-	return String(Math.floor(diff / DAY));
-};
-const d = time => {
-	const day = D(time);
-	if (day !== '11' && day.endsWith('1')) return `${day}st`;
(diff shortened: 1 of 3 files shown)


src/lib/util/Timestamp.js
```diff
@@ -153,114 +266,20 @@ class Timestamp {
...
-};
-
-/* eslint-enable id-length */
+/**
+ * The map tokens, mapped from the constants.
+ * @since 0.5.0
+ * @type {Map<string, Function>}
+ * @static
+ */
+Timestamp.mapTokens = new Map(Object.entries(TOKENS));
```
vladfrangu: What's the point of this being from Object.entries? It should also be `tokens` not `TOKENS` (latter is Record<string, number>)

```suggestion
Timestamp.mapTokens = new Map(tokens);
```

src/lib/util/Timestamp.js
```diff
@@ -153,114 +266,20 @@ class Timestamp {
...
-
-Timestamp.Z =
-Timestamp.ZZ = time => {
-	const offset = time.getTimezoneOffset();
-	return `${offset >= 0 ? '+' : '-'}${String(offset / -60).padStart(2, '0')}:${String(offset % 60).padStart(2, '0')}`;
-};
-
-/* eslint-enable id-length */
+/**
+ * The map tokens, mapped from the constants.
```
vladfrangu: This is inaccurate, see below

src/lib/util/Timestamp.js
kyranet: `TOKENS`' type is `Record<string, number>` and maps stuff like `Y: 4` (max length for the token, since it can read up to `YYYY`), `mapToken`'s type is `Map<string, number>`, aka the same.

`tokens`' type is `Map<string, (time: Date) => string>`, they're not applicable.

src/lib/util/Timestamp.js
kyranet: Yeah my brain went missing when I wrote this line, I don't know what to write there 😂

src/lib/util/Timestamp.js
vladfrangu: Three lines above... `* @type {Map<string, Function>}`
And how would that help us in any way, shape or form? We can already manually import the constants file and extract the tokens, while we cannot access the tokens Map you just implemented

kyranet: I guess I added it for consistency with the other classes, but I can 👢 it.

src/lib/util/Timestamp.js
```diff
@@ -1,5 +1,119 @@
 const { TIME: { SECOND, MINUTE, DAY, DAYS, MONTHS, TIMESTAMP: { TOKENS } } } = require('./constants');
 
+/* eslint-disable id-length, new-cap */
+
+// Dates
+
+const Y = time => String(time.getFullYear()).slice(0, 2);
```
bdistin: You are declaring a bunch of variables to use exactly once...

bdistin: You are declaring a bunch of variables to use exactly once...

kyranet: Expanded and inlined all the code 👍

src/lib/util/Timestamp.js
```diff
@@ -122,17 +196,18 @@ class Timestamp {
 		for (let i = 0; i < pattern.length; i++) {
 			let current = '';
 			const currentChar = pattern[i];
-			if (currentChar in TOKENS) {
+			if (tokenRepeatingCounts.has(currentChar)) {
```
bdistin: has + get antipattern

src/lib/util/Timestamp.js
```diff
@@ -1,5 +1,79 @@
...
+		return `${offset >= 0 ? '+' : '-'}${String(offset / -60).padStart(2, '0')}:${String(offset % 60).padStart(2, '0')}`;
+	}],
+	['ZZ', time => {
+		const offset = time.getTimezoneOffset();
+		return `${offset >= 0 ? '+' : '-'}${String(offset / -60).padStart(2, '0')}:${String(offset % 60).padStart(2, '0')}`;
+	}]
+]);
+/* eslint-enable max-len */
+
+const tokenRepeatingCounts = new Map(Object.entries(TOKENS));
```
bdistin: We only use TOKENS for this, if you want it to be a map, make it a map to start with instead of doubling up the memory footprint.

src/lib/util/Timestamp.js
kyranet: Addressed in 4d5be32

src/lib/util/Timestamp.js
kyranet: Addressed in 04ab830

PR #771, opened 2019-07-26
Command#cooldowns & Message#levelID  Props public

### LINKED ISSUE #770: Transferable Pieces Should Not Use Private Functions
### Describe the issue
In my opinion any piece that can be transferred should not use private functions. Once it is transferred then it becomes quite a pain to have to fix that. Either we also need to move the entire Class where those private methods are or we need to add ignores to the files which isn’t ideal either.

**Solution:**
Change any methods used in Pieces to be public.

```js
run(message: KlasaMessage, command: Command) {
	if (this.client.owners.has(message.author) || command.cooldown <= 0) return

	const existing = command.cooldowns.get(message.levelID)

	if (existing && existing.limited)
		throw message.language.get("INHIBITOR_COOLDOWN", this.secondsToFormattedString(Math.ceil(existing.remainingTime / 1000)), command.cooldownLevel !== "author")
}
```

image

### Further details

- **discord.js version**: master
- **node.js version**: 10.15
- **Klasa version**: settings branch
- [ ] I have tested the issue on latest master. Commit hash:

Skillz4Killz:
### Description of the PR
This PR fixes #770 

- Make private props into public

### Semver Classification

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -77,7 +77,7 @@ declare module 'klasa' {
 	}
 
 	export class KlasaMessage extends Message {
-		private levelID: Snowflake | null;
+		public levelID: Snowflake | null;
 		private prompter: CommandPrompt | null;
 		private _responses: KlasaMessage[];
 		private _patch(data: any): void;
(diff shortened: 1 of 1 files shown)


bdistin:
no

PR #775, opened 2019-08-03
Updated _parseCommand method

BannerBomb:
#### How it worked before this PR
If a user edits their command message and replaces its content with an invalid command and the command previously run contained subcommands it would error but if it didn't contain subcommands the client would run the same command a second time because the values were not redefined.

#### How it works now
Once a user runs a command then edits the contents that don't contain a valid prefix the properties are set to null since it would be an invalid command. Which tells the commandHandler that it's not a valid command to run.

-

### Semver Classification


bdistin:
This is fixed in 1.0.0-alpha with 754833e45aca191059435f446eb5b565db4fd6e7

PR #778, opened 2019-08-06
Added an options property to the Command class

BannerBomb:
### Description of the PR
I added an options property for commands. This will allow users to extend their commands further, by allowing them to pass extra options to commands then access them in for example 'Inhibitors'. The extra options can be accessed by doing `command.options['option name here']` the original built-in options can be accessed normally by doing something like `command.name` or `command.options.name`. This is also useful so users won't have to create a whole new Command class extending the existing klasa Command class, just to allow for more command options, with this they can continue to still use the original klasa command class for their commands while having the property to allow accessing extra options they want for their commands.

-

### Semver Classification


bdistin:
🤦‍♂

PR #822, opened 2019-09-30
Fix Broadcast Eval bug

### LINKED ISSUE #821: Broadcast Eval bug
### Describe the issue
In onceReady.js, the schedule is initialized before pieces, which means the coreSettings pieces are not disabled if sharding isn't used.
On rare occasions, the schedule will need to update a recurring task that missed it's time. This causes a db update, which attempts to update across shards thanks to the coreSettings events, which throws 
```js
Cannot read property 'broadcastEval' of null
```

```js

```

### Further details

- **discord.js version**: v12.0.0
- **node.js version**: v10-v12
- **Klasa version**: latest commit of settings
- [ ] I have tested the issue on latest master. Commit hash:

PyroTechniac:
### Description of the PR
Fixes #821 

-

### Semver Classification

### DIFF
--- a/src/events/coreSettingsCreate.js
+++ b/src/events/coreSettingsCreate.js
@@ -5,7 +5,7 @@ module.exports = class extends Event {
 
 	constructor(...args) {
 		super(...args, { event: 'settingsCreate' });
-		this.enabled = Boolean(this.client.shard);
+		if (!this.client.shard) this.disable();
 	}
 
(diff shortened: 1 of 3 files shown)


PyroTechniac:
Fixed in the latest commits

PyroTechniac:
Blocked as current solution has not fixed the bug.

PR #830, opened 2019-10-02
Fix ReactionHandler's permissions issues

Kizuru:
# Description of the PR
So, I'm using this class a lot, such as in my help command, music playlists, etc and I found out I would often get issues due to missing bot permissions. 
I thought it'd be nice to add the permission check inside the class itself rather than some custom checks I would add into Klasa's code (which is not a good solution since I have to upload separately the klasa module now rather than just `npm install`-ing it).

- Added permission check inside ReactionHandler for "MANAGE_MESSAGES" when deleting a reaction from a message.

### Semver Classification

### DIFF
--- a/src/lib/util/ReactionHandler.js
+++ b/src/lib/util/ReactionHandler.js
@@ -111,7 +111,8 @@ class ReactionHandler extends ReactionCollector {
 		else return this.stop();
 
 		this.on('collect', (reaction, user) => {
-			reaction.users.remove(user);
+			const clientPermissions = reaction.message.channel.permissionsFor(reaction.message.client);
+			if (clientPermissions.has("MANAGE_MESSAGES")) reaction.users.remove(user);
 			this[this.methodMap.get(reaction.emoji.id || reaction.emoji.name)](user);
 		});
 		this.on('end', () => {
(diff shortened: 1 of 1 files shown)


src/lib/util/ReactionHandler.js
```diff
@@ -111,7 +111,8 @@ class ReactionHandler extends ReactionCollector {
 		else return this.stop();
 
 		this.on('collect', (reaction, user) => {
-			reaction.users.remove(user);
+			const clientPermissions = reaction.message.channel.permissionsFor(reaction.message.client);
+			if (clientPermissions.has("MANAGE_MESSAGES")) reaction.users.remove(user);
```
kyranet: To fix the ESLint error:
```suggestion
			if (clientPermissions.has('MANAGE_MESSAGES')) reaction.users.remove(user);
```

Kizuru:
Oops my bad I had forgotten js linters won't like double quotes for these kind of purpose.

bdistin:
This has been suggested many times and denied many times more. You are given the tools to guard permissions before creating a reaction handler. Everyone doesnt need to suffer with extra slow code because the few are too lazy to use those tools to ensure reaction handlers can be used. This is not the only place manage messages is needed btw.

PR #852, opened 2019-10-24
Fix(Extendable): Fix TS decorators breaking Extendables

PyroTechniac:
### Description of the PR
When an extendable is decorated, Object.getOwnPropertyKeys() no longer returns the property names, not even the built in ones (constructor, prototype, name, length)

The fix in action can be viewed [here](https://gist.github.com/PyroTechniac/ebf07fa3c1c458810563bbf0957295d9)

-

### Semver Classification

### DIFF
--- a/src/lib/structures/Extendable.js
+++ b/src/lib/structures/Extendable.js
@@ -31,8 +31,10 @@ class Extendable extends Piece {
 		super(store, file, directory, options);
 
 		const staticPropertyNames = Object.getOwnPropertyNames(Object.getPrototypeOf(this.constructor))
+			.concat(Object.getOwnPropertyNames(this.constructor)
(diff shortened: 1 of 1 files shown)


PyroTechniac:
It should be noted that if a class being passed to appliesTo has been decorated, then Object.getOwnPropertyDescriptor will return undefined unless it is also made with this fix, which I haven't included but can if needed

tech6hutch:
Does this break "normal" extendables?

PyroTechniac:
@tech6hutch fixed in 866153d151e9f30fb11d5461d31305f9e64d5d3a

PyroTechniac:
As per discord discussion, this shouldn't be merged, and should instead wait for decorators to land in JS

PR #894, opened 2020-01-01
Remove client from CommandUsage constructor, and remove some old typings

PyroTechniac:
### Description of the PR
Removes the client from the CommandUsage constructor (as it's present on the Command instance) and cleans up some typings.

-

### Semver Classification


bdistin:
1. This is 2 unrelated PRs, and should not be combined.
2. The main objective doesn't actually clean up or make anything easier really. (CommandUsage is only used internally)
3. Now CommandUsage would not properly extend Usage (in terms of strictness). Yes, this can be done in JS, but it's not appreciated when done in TS.

PR #911, opened 2020-01-28
Update CHANGELOG.md

TheDrone7:
# Description of the PR
Modified `this.client.options.totalShardCount` to `this.client.options.shardCount` as per discord.js master branch documentation.

- Fixed the `stats` command as which didn't show the actual shard count due because of non-existent property usage.

### Semver Classification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -291,6 +291,7 @@ NOTE: For the contributors, you add new entries to this document following this 
 
 ### Fixed
 
+- [] Fixed shard count in the stats command.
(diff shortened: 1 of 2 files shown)


tjrgg:
Is there meant to be other changes in this PR? I'm only seeing an update to the changelog and nothing more.

TheDrone7:
> Is there meant to be other changes in this PR? I'm only seeing an update to the changelog and nothing more.

The language `en-US` was also supposed to be modified. I think I probably forgot committing that. I've committed that too now. As it currently shows `1 / undefiend` when using the `stats` command because there is minor change in discord.js

bdistin:
This is incompatible with the full rewrite, closing this.

PR #931, opened 2020-02-22
Throw error instead of emit error for fixing unexpected type

yuta0801:
### Description of the PR

Store#set() is expected to return `?Piece`, but may return boolean because of returns client.emit() directly.
EventEmitter.emit() returns true if the event had listeners, false otherwise.
See: https://nodejs.org/api/events.html#events_emitter_emit_eventname_args

Code affected by this behavior: https://github.com/dirigeants/klasa/blob/ea059931fad07bdea870620181de453efe816646/src/lib/structures/base/AliasStore.js#L50-L52

- Fixed unexpected return type of Store#set() becouse of returning the client.emit() directly

### Semver Classification

### DIFF
--- a/src/lib/structures/base/Store.js
+++ b/src/lib/structures/base/Store.js
@@ -133,13 +133,10 @@ class Store extends Collection {
 	 * Sets up a piece in our store.
 	 * @since 0.0.1
 	 * @param {Piece} piece The piece we are setting up
-	 * @returns {?Piece}
+	 * @returns {Piece}
 	 */
 	set(piece) {
-		if (!(piece instanceof this.holds)) {
-			this.client.emit('error', `Only ${this} may be stored in this Store.`);
-			return null;
-		}
+		if (!(piece instanceof this.holds)) throw new Error(`Only ${this} may be stored in this Store.`);
 		const existing = this.get(piece.name);
 		if (existing) this.delete(existing);
 		else if (this.client.listenerCount('pieceLoaded')) this.client.emit('pieceLoaded', piece);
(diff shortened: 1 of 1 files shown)


bdistin:
actually it should just throw a full error.

yuta0801:
I think it rarely happens unexpected behavior, but it returns true when someone handling `error` event of client (`.emit()` only throws an error when no `error` event handler)
And at this time, pass the wrong Peace to `AliasStore#set()`, TypeError will occur because of `(true).aliases` is executed

PyroTechniac:
That will cause more confusion and be harder for both users and framework developers to find the source of the error.

This is definitely a bug in the framework but I agree that it should just throw an error, as the client will stop loading if an error occurs (which it should)

yuta0801:
Should I change to throw an error instead of explicitly returning null after emitting the error?

bdistin:
yeah

src/lib/structures/base/Store.js
```diff
@@ -133,13 +133,10 @@ class Store extends Collection {
...
 	 * @param {Piece} piece The piece we are setting up
-	 * @returns {?Piece}
+	 * @returns {Piece}
 	 */
 	set(piece) {
-		if (!(piece instanceof this.holds)) {
-			this.client.emit('error', `Only ${this} may be stored in this Store.`);
-			return null;
-		}
+		if (!(piece instanceof this.holds)) throw new Error(`Only ${this} may be stored in this Store.`);
```
yuta0801: Is TypeError better than just Error?

PyroTechniac: TypeError would be a more accurate error to use, yes

